PR #8, opened 2021-11-25
Multiple cancels

epsilonsized:
* Multiple cancel method which takes in an array of cancel arg objects
* Anything over 4 cancels will be split into multiple txs


src/client.ts
```diff
@@ -641,3 +673,10 @@ export class Client {
     }
   }
 }
+
+export interface CancelArgs{
```
kelvinlau20100, line 677: can you move this to types.ts instead?

epsilonsized: thanks will do

kelvinlau20100 (approved):
lgtm

PR #9, opened 2021-11-26
Insurance fund changes

kelvinlau20100:


src/client.ts
```diff
@@ -64,6 +71,22 @@ export class Client {
   }
   private _marginAccountAddress: PublicKey;
 
+  /**
+   * CLient insurance vault deposit account to track how much they deposited / are allowed to withdraw
+   */
+  public get insuranceDepositAccount(): InsuranceDepositAccount | null {
```
xjcaa, line 77: Hm - The only thing here is do we need this in client - I feel like it could be split out from what is a `Client`?  it's almost as if normal clients never should care about this.

src/client.ts
```diff
@@ -311,21 +353,7 @@ export class Client {
...
-        `Found user USDC associated token account ${this._usdcAccountAddress.toString()}. Balance = $${utils.getReadableAmount(
-          tokenAccountInfo.amount.toNumber()
-        )}.`
-      );
-    } catch (e) {
-      throw Error(
-        "User has no USDC associated token account. Please create one and deposit USDC."
-      );
-    }
+    this.usdcAccountCheck();
```
xjcaa, line 356: await

src/client.ts
xjcaa: Alternatively, its fine to keep in - you add a function as well that checks if a client is whitelisted to deposit.

src/client.ts
```diff
@@ -657,6 +751,24 @@ export class Client {
     this._positions = positions;
   }
 
+  private async usdcAccountCheck() {
```
xjcaa, line 754: nice

src/client.ts
```diff
@@ -223,6 +280,24 @@ export class Client {
       console.log("User does not have a margin account.");
     }
 
+    try {
+      client._whitelistInsuranceAccount =
+        (await client._program.account.whitelistInsuranceAccount.fetch(
+          client._whitelistInsuranceAccountAddress
+        )) as WhitelistInsuranceAccount;
+    } catch (e) {
+      console.log("User does not have a whitelist insurance account.");
```
xjcaa, line 289: Hmm I feel like this is noise - as most users will not have this or a insurance vault account.

xjcaa: Also if they don't have a whitelist insurance account, they most definitely won't have a insurance deposit account - so put it together

xjcaa: Feels like you can use whitelistInsuranceCheck here? to reduce duplicated code. or consolidate it into one?

src/exchange.ts
```diff
@@ -663,6 +661,33 @@ strikeInitializationThresholdSeconds=${params.strikeInitializationThresholdSecon
     );
   }
 
+  public async rebalanceInsuranceVault(remainingAccounts: any[]) {
```
xjcaa, line 664: I think can call these MarginAccounts? to be more clear

xjcaa:
Can we add a function in `Exchange` to whitelist a user to deposit to insurance fund?

xjcaa (approved):
lgtm

PR #24, opened 2021-12-09
Platform fees

kelvinlau20100:
associated sdk changes for PR https://github.com/zetamarkets/zeta-options/pull/322

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -123,6 +123,15 @@ export class Client {
   private _pendingUpdate: boolean;
 
   /**
+   * whitelist trading fees account.
+   */
+  private _whitelistTradingFeesAddress: PublicKey;
+
+  /**
+   * Whitelited for trading fees.
+   */
+  private _whitelistedTradingFees: boolean;
+  /**
    * Polling interval.
    */
   public get pollInterval(): number {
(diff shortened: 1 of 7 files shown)


src/client.ts
```diff
@@ -234,6 +243,24 @@ export class Client {
       client._pendingUpdate = true;
     }
 
+    let [whitelistTradingFeesAddress, _whitelistTradingFeesNonce] =
+      await utils.getUserWhitelistTradingFeesAccount(
+        Exchange.programId,
+        wallet.publicKey
+      );
+    client._whitelistTradingFeesAddress = whitelistTradingFeesAddress;
+
```
xjcaa, line 252: I think what you can do here is just - 

1. Get PDA
2. Check existence
3. If exist, set whitelistTradingFeesAddress to pda, else leave as undefined and add log, else dont log about whitelist for non whitelisting.

xjcaa: Then you dont need the trading fees boolean.

src/client.ts
```diff
@@ -374,15 +401,29 @@ export class Client {
...
-      marketIndex,
-      price,
-      size,
-      side,
-      this.marginAccountAddress,
-      this.publicKey,
-      openOrdersPda
-    );
+    let orderIx;
+    if (this._whitelistedTradingFees) {
```
xjcaa, line 405: Then this can be - 
```
if (this._whitelistedTradingFees !== undefined) ...
```

src/client.ts
```diff
@@ -374,15 +401,29 @@ export class Client {
...
+      orderIx = await placeOrderIx(
+        marketIndex,
+        price,
+        size,
+        side,
+        this.marginAccountAddress,
+        this.publicKey,
+        openOrdersPda,
+        this._whitelistTradingFeesAddress
```
xjcaa, line 414: If `this._whitelistTradingFeesAddress` == undefined, you could just directly pass it in without needing an if else and then check if it's undefined in the program-instruction `placeOrderIx` function instead and it'd be cleaner.

src/client.ts
kelvinlau20100: yeah this makes a lot of sense

xjcaa (approved):

PR #30, opened 2021-12-11
Typedoc documentation

0xtristan:
- Added proper TS docs
Looking at other repos (solana/web3.js and raydium) they have nicely formatted SDK documentation. Turns out this is all simple and handled by a nice package called Typedoc. Adding this to our docs so they are a bit more professional. Also includes github action to publish to Github pages.

- Updated license in package.json
- Changed package manager from npm to yarn


0xtristan:
IMO we should also add a readme on how to contribute etc.

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,44 @@
...
+  push:
+    branches: 
+    - main
+    - release/*
+
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+jobs:
+#   publish_npm:
```
xjcaa, line 13: delete?

xjcaa:
> IMO we should also add a readme on how to contribute etc.

Just add a PR?

.github/workflows/deploy.yml
0xtristan: Yeah I was leaving in in case we decide to automate it later

PR #31, opened 2021-12-12
Margin parameters onchain.

xjcaa:


src/client.ts
```diff
@@ -29,6 +29,7 @@ import {
 
 import { Position, Order, Side } from "./types";
 import { EventType } from "./events";
+import * as constants from "./constants";
```
kelvinlau20100, line 32: Unneeded import?

kelvinlau20100 (approved):
LGTM, one nit

PR #34, opened 2021-12-16
Platform halt

xjcaa:


src/market.ts
```diff
@@ -599,7 +598,5 @@ export class Market {
         await processTransaction(Exchange.provider, tx);
       })
     );
-
-    await sleep(1000);
```
xjcaa, line 603: @kelvinlau20100 do you remember why we had this sleep lol

src/exchange.ts
```diff
@@ -1016,6 +940,90 @@ insuranceVaultLiquidationPercentage=${params.insuranceVaultLiquidationPercentage
...
+    this.assertHalted();
+    await Promise.all(
+      this._markets.markets.map(async (market) => {
+        await market.cancelAllOrdersHalted();
+      })
+    );
+  }
+
+  public async cleanZetaMarketsHalted() {
+    this.assertHalted();
```
0xtristan, line 995: Is this not already asserted on the rust side?

0xtristan (approved):
Yeet, looks great

src/exchange.ts
xjcaa: Rather it fail in the client too.

PR #35, opened 2021-12-17
Lot size

kelvinlau20100:


kelvinlau20100:
one thing that could be good is to make util functions for converting to native position size and back

xjcaa:
> 

Can you add this

src/client.ts
```diff
@@ -396,7 +397,7 @@ export class Client {
     let orderIx = await placeOrderIx(
       marketIndex,
       price,
-      size,
+      size * Math.pow(10, POSITION_PRECISION),
```
xjcaa, line 400: Can we pass in the raw number - like 1000 = 1 lot

xjcaa: And do the conversion on the outside

src/risk.ts
```diff
@@ -229,7 +247,9 @@ export function calculateLiquidationPrice(
     return 0;
   }
   let availableBalance = accountBalance - marginRequirement + unrealizedPnl;
-  return markPrice - availableBalance / position;
+  return (
+    markPrice - availableBalance / (position / Math.pow(10, POSITION_PRECISION))
```
xjcaa, line 251: Can we just make 1_000 a constant.

src/client.ts
```diff
@@ -610,7 +611,7 @@ export class Client {
       this._marginAccountAddress,
       market,
       liquidatedMarginAccount,
-      size
+      size * Math.pow(10, POSITION_PRECISION)
```
xjcaa, line 614: This needs to be fixed

PR #85, opened 2022-01-29
Initial margin changes.

xjcaa:


examples/liquidator-example/src/liquidator-utils.ts
```diff
@@ -46,17 +46,16 @@ export async function findLiquidatableAccounts(
...
-      if (adjustedAvailableBalance >= 0) {
+      // Therefore can just use availableBalanceLiquidation which assumes 0 open orders.
+      if (marginAccountState.availableBalanceLiquidation >= 0) {
         return;
       }
       console.log(
         `[LIQUIDATABLE ACCOUNT] [ACCOUNT] ${account.publicKey.toString()} [BALANCE] ${
           marginAccountState.balance
-        } [ADJUSTED AVAILABLE BALANCE] ${adjustedAvailableBalance}`
+        } [AVAILABLE BALANCE MAINTENANCE] ${
```
epsilonsized, line 56: preference to keep logging terminology the same as the sdk terminology

src/risk.ts
```diff
@@ -135,32 +131,50 @@ export class RiskCalculator {
...
+
+      let longLots = convertNativeLotSizeToDecimal(
+        position.openingOrders[0].toNumber()
+      );
+
+      let shortLots = convertNativeLotSizeToDecimal(
+        position.openingOrders[1].toNumber()
+      );
+
+      if (position.position.toNumber() >= 0) {
```
epsilonsized, line 148: this can be an > sign innit

epsilonsized, line 148: why is this a >= instead of a >? makes it inconsistent with the program code although I understand no impact. More for my understanding

src/risk.ts
```diff
@@ -205,15 +208,16 @@ export class RiskCalculator {
     let unrealizedPnl = this.calculateUnrealizedPnl(marginAccount);
     let initialMargin = this.calculateTotalInitialMargin(marginAccount);
     let maintenanceMargin = this.calculateTotalMaintenanceMargin(marginAccount);
-    let totalMargin = initialMargin + maintenanceMargin;
-    let availableBalance: number = balance + unrealizedPnl - totalMargin;
+    let availableBalance: number = balance + unrealizedPnl - initialMargin;
```
epsilonsized, line 211: @xjcaa @kelvinlau20100 thoughts on terminology here?

Would availableBalanceInitial and availableBalanceMaintenance just make life easier / more sense across the board? 

Dislike availableBalance as a generic name

src/types.ts
```diff
@@ -140,9 +140,9 @@ export interface MarginAccountState {
   balance: number;
   initialMargin: number;
   maintenanceMargin: number;
-  totalMargin: number;
   unrealizedPnl: number;
   availableBalance: number;
+  availableBalanceLiquidation: number;
```
epsilonsized, line 145: similar comment to above. 

Think changing the terminology to make the distinction to availableBalanceInitial and availableBalanceMaintenance forces clients to recognise the difference is less lost in the noise(?).

PR #93, opened 2022-02-09
Order Types and Close Accounts SDK changes

kelvinlau20100:


src/client.ts
```diff
@@ -651,7 +851,111 @@ export class Client {
...
+        throw Error("User has no open orders account for this market!");
+      }
+
+      const [vaultOwner, _vaultSignerNonce] =
+        await utils.getSerumVaultOwnerAndNonce(
+          market,
+          DEX_PID[Exchange.network]
+        );
+
+      settleIxs.push(
```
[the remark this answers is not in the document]
kelvinlau20100: updated

src/client.ts
```diff
@@ -626,6 +776,55 @@ export class Client {
...
+        newOrderSide,
+        newOrderType,
+        newOrderClientOrderId,
+        this.marginAccountAddress,
+        this.publicKey,
+        this._openOrdersAccounts[marketIndex],
+        this._whitelistTradingFeesAddress
+      )
+    );
+    ixs.forEach((ix) => tx.add(ix));
```
xjcaa, line 824: should we remove `forEach`?

src/client.ts
```diff
@@ -580,6 +679,57 @@ export class Client {
...
+      cancelOrderIx(
+        marketIndex,
+        this.publicKey,
+        this._marginAccountAddress,
+        this._openOrdersAccounts[marketIndex],
+        orderId,
+        cancelSide
+      )
+    );
+    ixs.push(
```
xjcaa, line 715: can't you just add directly to the tx.

src/client.ts
xjcaa: there's only two ix's y ou can just add directly.

xjcaa (approved):

PR #102, opened 2022-04-25
Spread accounts

xjcaa:
- [x] - Add function to simulate user spread account requirements post movement.
- [x] - Add function to simulate user's margin account savings and net balance.
- [x] - Migrate to V3 place order


src/client.ts
```diff
@@ -436,6 +522,37 @@ export class Client {
...
+    );
+    tx.add(
+      closeSpreadAccountIx(
+        Exchange.zetaGroupAddress,
+        this._spreadAccountAddress,
+        this.publicKey
+      )
+    );
+    let txId = await utils.processTransaction(this._provider, tx);
+    console.log(`[CLOSE SPREAD ACCOUNT] Transaction: ${txId}`);
```
kelvinlau20100, line 550: no need for this log imo

kelvinlau20100: actually there's a bunch of random logs like on closeMarginAccount and deposit where we log the txId? Can we just get rid of them

kelvinlau20100 (approved):
LGTM, just 1 nit

PR #103, opened 2022-04-26
Autogenerated changes due to U64 liquidation size

filipzeta:
(Auto-generated from changes in https://github.com/zetamarkets/zeta-options/pull/415)

Liquidate size has been bumped to u64 in the zeta-options program, and the changes here are autogenerated based on that. Note that the typescript 'number' type is actually limited to ~u53 in practice 😊

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ Version changes are pinned to SDK releases.
 
 ## [Unreleased]
 
-- program: Change liquidate() size argument from u32 to u64.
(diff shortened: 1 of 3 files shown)


xjcaa:
To fix the number 53 bit shenanigans, we could also just take BN. But don't think anyone is trading that big yet :p

xjcaa (approved):
LGTM - Can you add the PR to the changelog.

filipzeta:
> 

yeah i kept `number` just for consistency, but can bump it up to BN later if necessary :)

PR #115, opened 2022-05-11
Add oracle timestamp

filipzeta:
Manually tested this one by running the SDK locally. Example output when printing getPrice() and getPriceAge() in a loop:

_**Note: These don't include the lastUpdatedSlot because I added that last minute and don't want to remake all the logs :)**_
`{ feed: 'SOL/USD', price: 65.78932583, lastUpdatedTime: 1652248783 }`
`2.322999954223633`
`{ feed: 'SOL/USD', price: 65.78932583, lastUpdatedTime: 1652248785 }`
`3.3259999752044678`
I turn off my wifi here
`{ feed: 'SOL/USD', price: 65.78932583, lastUpdatedTime: 1652248785 }`
`6.332000017166138`
`{ feed: 'SOL/USD', price: 65.78932583, lastUpdatedTime: 1652248785 }`
`9.335999965667725`
`{ feed: 'SOL/USD', price: 65.78932583, lastUpdatedTime: 1652248785 }`
`12.348000049591064`
`{ feed: 'SOL/USD', price: 65.78932583, lastUpdatedTime: 1652248785 }`
`15.348999977111816`
`{ feed: 'SOL/USD', price: 65.78932583, lastUpdatedTime: 1652248785 }`
`18.34999990463257`
`{ feed: 'SOL/USD', price: 65.78932583, lastUpdatedTime: 1652248785 }`
`21.354000091552734`
I turn my wifi back on here
`{ feed: 'SOL/USD', price: 65.87564083, lastUpdatedTime: 1652248790 }`
`19.358999967575073`
`{ feed: 'SOL/USD', price: 65.87791666, lastUpdatedTime: 1652248802 }`
`10.358999967575073`
`{ feed: 'SOL/USD', price: 65.96525253, lastUpdatedTime: 1652248813 }`
`2.364000082015991`
`{ feed: 'SOL/USD', price: 65.98317166, lastUpdatedTime: 1652248816 }`
`2.365000009536743`

PollPrice() was tested by just commenting out the internal price update in the oracle code, to simulate that we're not getting any prices. Then we call sleep(3000) + getPrice() + getPriceAge():
`{ feed: 'SOL/USD', price: 65.9148975, lastUpdatedTime: 1652248960 }`
`4.359999895095825`
Then  pollPrice() + getPrice() + getPriceAge():
`{ feed: 'SOL/USD', price: 65.87878, lastUpdatedTime: 1652248966 }`
`2.1630001068115234`
sleep(3000) + getPrice() + getPriceAge():
`{ feed: 'SOL/USD', price: 65.87878, lastUpdatedTime: 1652248966 }`
`5.168999910354614`
pollPrice() + getPrice() + getPriceAge():
`{ feed: 'SOL/USD', price: 65.92875000000001, lastUpdatedTime: 1652248974 }`
`0.9619998931884766`


src/oracle.ts
```diff
@@ -51,13 +76,10 @@ export class Oracle {
         priceAddress,
         (accountInfo: AccountInfo<Buffer>, _context: Context) => {
           let priceData = parsePythData(accountInfo.data);
-          let currPrice = this._data.get(feed);
```
filipzeta, line 54: If prices are stale for a while but then recover to the same price, `currPrice.price === priceData.price`. Therefore we won't update the price until oracle price moves again.

Thoughts on removing this?

filipzeta:
Do we want to do anything with the Pyth `Last Slot` or `Valid Slot`? Technically Pyth could be stale but the `Exchange.clockTimestamp` would still update, but on the other hand that is very unlikely and also slot numbers are probably not useful to the end user unless we convert it back to a timestamp

CHANGELOG.md
```diff
@@ -5,6 +5,8 @@ Version changes are pinned to SDK releases.
 
 ## [Unreleased]
 
+- oracle: Add lastUpdatedTime and pollPrice() to Oracle. ([#115](https://github.com/zetamarkets/sdk/pull/115))
```
xjcaa, line 8: nit: Make it something like;

- oracle: `OraclePrice` now contains `lastUpdatedTime` which is seconds since linux epoch. `getPriceAge` can be called to get the time between now and lastUpdatedTime.
- oracle: Now has explicit `pollPrice` to fetch the latest oracle price manually.

src/oracle.ts
```diff
@@ -30,13 +30,38 @@ export class Oracle {
...
+
   // Allows fetching of any pyth oracle price.
   public async fetchPrice(oracleKey: PublicKey): Promise<number> {
     let accountInfo = await this._connection.getAccountInfo(oracleKey);
     let priceData = parsePythData(accountInfo.data);
     return priceData.price;
   }
 
+  // Fetch and update an oracle price manually
+  public async pollPrice(feed: string, triggerCallback = true) {
```
xjcaa, line 45: I think this should also just return the oracleData.

src/oracle.ts
xjcaa: Seems fine - I think it was added with intent to reduce spam.

src/oracle.ts
```diff
@@ -68,13 +90,7 @@ export class Oracle {
...
       // TODO set this so localnet has data for the oracle
       // Remove once there is an oracle simulator.
-      let accountInfo = await this._connection.getAccountInfo(priceAddress);
-      let priceData = parsePythData(accountInfo.data);
-      let oracleData = {
-        feed,
-        price: priceData.price,
-      };
-      this._data.set(feed, oracleData);
+      this.pollPrice(feed, false);
```
xjcaa, line 93: needs an `await`

src/oracle.ts
```diff
@@ -88,4 +104,5 @@ export class Oracle {
 export interface OraclePrice {
   feed: string;
   price: number;
+  lastUpdatedTime: number;
```
xjcaa, line 107: I think while we're here, may as well add lastUpdatedSlot; you can get this from the subscription callback `context` which is unused.

src/oracle.ts
```diff
@@ -30,13 +30,40 @@ export class Oracle {
...
+
   // Allows fetching of any pyth oracle price.
   public async fetchPrice(oracleKey: PublicKey): Promise<number> {
     let accountInfo = await this._connection.getAccountInfo(oracleKey);
     let priceData = parsePythData(accountInfo.data);
     return priceData.price;
   }
 
+  // Fetch and update an oracle price manually
+  public async pollPrice(feed: string, triggerCallback = true) {
```
xjcaa, line 45: Can type this to `: OraclePrice` as the return value.

xjcaa:
great testing fwiw

xjcaa (approved):

PR #124, opened 2022-05-30
Multi underlying sdk v2

filipzeta:


examples/basic/bot-key.json
```diff
@@ -0,0 +1 @@
+[184,213,164,163,252,35,61,169,234,13,28,115,74,219,126,234,48,80,213,130,122,21,19,22,249,153,124,0,146,37,68,25,255,121,38,82,221,6,221,21,124,88,177,245,164,104,149,206,150,150,3,233,116,106,5,150,34,41,121,151,13,230,203,143]
```
filipzeta, line 1: devnet key dont hack me pls

src/client.ts
```diff
@@ -1,1033 +1,238 @@
...
+import { exchange as Exchange } from "./exchange";
+import * as instructions from "./program-instructions";
+import * as fs from "fs";
+import * as os from "os";
 
 export class Client {
-  /**
-   * Returns the user wallet public key.
-   */
-  public get publicKey(): PublicKey {
```
xjcaa, line 71: This should exist at the higher level. Same with provider, program.

src/client.ts
```diff
@@ -1,1033 +1,238 @@
...
-   */
-  public get spreadAccountAddress(): PublicKey {
-    return this._spreadAccountAddress;
-  }
-  private _spreadAccountAddress: PublicKey;
-
-  /**
-   * Client usdc account address.
-   */
-  public get usdcAccountAddress(): PublicKey {
```
xjcaa, line 123: This is global too

src/client.ts
```diff
@@ -1,1033 +1,238 @@
...
-   */
-  private _pendingUpdateSlot: number = 0;
-
-  /**
-   * whitelist deposit account.
-   */
-  public get whitelistDepositAddress(): PublicKey | undefined {
-    return this._whitelistDepositAddress;
+  private constructor() {
+    this._subClients = new Map();
```
xjcaa, line 49: I wonder if using a Map is better than just indexing in via Asset into a list

src/client.ts
```diff
@@ -1,1033 +1,238 @@
...
-    tag: String = DEFAULT_ORDER_TAG
-  ): Promise<TransactionSignature> {
-    let tx = new Transaction();
+  public async cancelAllOrdersAllAssets() {
+    let txIdMap = new Map();
 
-    let marketIndex = Exchange.markets.getMarketIndex(market);
+    await Promise.all(
+      this.getAllSubClients().map(async (subClient) => {
+        let txId = await subClient.cancelAllOrders();
```
xjcaa, line 153: nit: `txIds`?

src/client.ts
```diff
@@ -1,1033 +1,238 @@
...
+        await subClient.updateState(fetch, force);
+      })
     );
+  }
 
-    tx.add(orderIx);
-
-    let txId = await utils.processTransaction(this._provider, tx);
-    this._openOrdersAccounts[marketIndex] = openOrdersPda;
+  public async cancelAllOrders(asset: Asset) {
```
xjcaa, line 140: default to null, cancel all 

only cancel specific if passed in?

src/client.ts
```diff
@@ -1,1033 +1,238 @@
...
-      price,
-      size,
-      side,
-      orderType,
-      clientOrderId,
-      this.marginAccountAddress,
-      this.publicKey,
-      openOrdersPda,
-      this._whitelistTradingFeesAddress
+  public async updateStateAllAssets(fetch = true, force = false) {
```
xjcaa, line 132: I think with all these `AllAssets` postfix, we can make it a flag instead defaulting to all, and only specific if an asset is passed in

src/client.ts
```diff
@@ -1,1033 +1,238 @@
...
-   */
-  public async placeOrderAndLockPosition(
-    market: PublicKey,
-    price: number,
-    size: number,
-    side: Side,
-    tag: String = DEFAULT_ORDER_TAG
-  ): Promise<TransactionSignature> {
-    let tx = new Transaction();
+  public async cancelAllOrdersAllAssets() {
```
xjcaa, line 148: I kinda dislike returning a Map, seems too convoluted. Maybe just return an array of tx signatures?

Also can you specify return type as well;

src/errors.ts
```diff
@@ -144,7 +144,7 @@ export class NativeError extends Error {
       },
       _programErrorStack: ProgramErrorStack { stack: [ [PublicKey] ] }
     }
- * Anchor error is rich but in information but breaks the assumptions on errors by existing clients.
+ * Anchor error is rich but in information but breaks the assumptions on errors by existing subClients.
```
xjcaa, line 147: accidental search and replace? :p

src/exchange.ts
```diff
@@ -35,185 +41,13 @@ export class Exchange {
   }
   private _isInitialized: boolean = false;
 
-  /**
-   * The solana network being used.
-   */
-  public get network(): Network {
```
xjcaa, line 41: network, program , provider, state, clock, stateaddress, serumAuthority, mintAuthority should be global

src/exchange.ts
```diff
@@ -231,922 +65,306 @@ export class Exchange {
...
-    // Set callbacks.
-    exchange.subscribeZetaGroup(callback);
-    exchange.subscribeGreeks(callback);
-
-    await exchange.subscribeClock(callback);
-
-    exchange._isInitialized = true;
+    await Promise.all(
+      assets.map(async (asset) => {
+        this._riskCalculator = new RiskCalculator(assets);
```
xjcaa, line 100: Doesn't need to be in the promise all

src/exchange.ts
```diff
@@ -231,922 +65,306 @@ export class Exchange {
...
+            connection,
+            opts,
+            wallet,
+            throttleMs,
+            callback
+          );
 
-    console.log(
-      `Exchange loaded @ ${new Date(exchange.clockTimestamp * 1000)}`
+        this._oracle = new Oracle(network, connection);
```
xjcaa, line 115: same as risk calc

src/exchange.ts
```diff
@@ -231,922 +65,306 @@ export class Exchange {
...
+        await this.subscribeOracle(assets, callback);
+        this._isInitialized = true;
```
xjcaa, line 117: should be set once

src/exchange.ts
```diff
@@ -231,922 +65,306 @@ export class Exchange {
...
-  ) {
-    let tx = new Transaction().add(
-      instructions.updateMarginParametersIx(
-        args,
-        this._provider.wallet.publicKey
-      )
-    );
-    await utils.processTransaction(this._provider, tx);
-    await this.updateZetaGroup();
+  public getSubExchange(asset: Asset) {
```
xjcaa, line 172: add return types for all functions that return sth

src/exchange.ts
```diff
@@ -231,922 +65,306 @@ export class Exchange {
...
-      );
-    }
-    let tx = new Transaction().add(
-      instructions.updateVolatilityNodesIx(
-        nodes,
-        this._provider.wallet.publicKey
-      )
-    );
-    await utils.processTransaction(this._provider, tx);
+  public getAllSubExchanges() {
```
xjcaa, line 176: unsure of use case for this tbh

src/exchange.ts
```diff
@@ -231,922 +65,306 @@ export class Exchange {
...
-    await utils.processTransaction(this._provider, tx);
-  }
-
-  /**
-   * Retreat volatility surface and interest rates for an expiry index.
-   */
-  public async retreatMarketNodes(expiryIndex: number) {
-    let tx = new Transaction().add(
-      instructions.retreatMarketNodesIx(expiryIndex)
+  public async updateAllAssetsOrderbooks(index: number) {
```
xjcaa, line 205: Same deal w client

xjcaa: Also when would you ever update all assets for one index

src/exchange.ts
```diff
@@ -231,922 +65,306 @@ export class Exchange {
...
+      })
     );
-    await utils.processTransaction(this._provider, tx);
   }
 
-  public assertInitialized() {
-    if (!this.isInitialized) {
-      throw "Exchange uninitialized";
-    }
+  public getOrderbook(asset: Asset, index: number) {
```
xjcaa, line 213: nit: add return types

src/exchange.ts
```diff
@@ -231,922 +65,306 @@ export class Exchange {
...
-      throw Error("Cannot subscribe greeks. ZetaGroup is null.");
-    }
+  public async updateVolatilityNodes(asset: Asset, nodes: Array<anchor.BN>) {
+    await this.getSubExchange(asset).updateVolatilityNodes(nodes);
+  }
 
-    let eventEmitter = this._program.account.greeks.subscribe(
-      this._zetaGroup.greeks,
-      this._provider.connection.commitment
+  public async initializeAllAssetsZetaMarkets() {
```
xjcaa, line 239: delet

src/exchange.ts
```diff
@@ -231,922 +65,306 @@ export class Exchange {
...
-   */
-  public async getDepositLimit() {
-    return utils.convertNativeBNToDecimal(this.state.nativeDepositLimit);
+  public async whitelistUserForDeposit(asset: Asset, user: PublicKey) {
+    await this.getSubExchange(asset).whitelistUserForDeposit(user);
   }
 
-  public addProgramSubscriptionId(id: number) {
-    this._programSubscriptionIds.push(id);
+  public async whitelistUserForTradingFees(asset: Asset, user: PublicKey) {
```
xjcaa, line 279: Whitelisting is global aka whitelist account seeds are done on a userkey basis with no knowledge of the underlying

src/exchange.ts
```diff
@@ -231,922 +65,306 @@ export class Exchange {
...
-      optionDynamicPercentageShortMaintenance: utils.convertNativeBNToDecimal(
-        this.zetaGroup.marginParameters.optionDynamicPercentageShortMaintenance,
-        constants.MARGIN_PRECISION
-      ),
-      optionShortPutCapPercentage: utils.convertNativeBNToDecimal(
-        this.zetaGroup.marginParameters.optionShortPutCapPercentage,
-        constants.MARGIN_PRECISION
-      ),
-    };
+  public async getDepositLimit(asset: Asset) {
```
xjcaa, line 287: Should be global

src/exchange.ts
```diff
@@ -231,922 +65,306 @@ export class Exchange {
...
+    await this.getSubExchange(asset).updateInterestRate(args);
   }
 
-  public async updateInterestRate(args: instructions.UpdateInterestRateArgs) {
-    let tx = new Transaction().add(
-      instructions.updateInterestRateIx(args, this._provider.wallet.publicKey)
+  public async close() {
+    await Promise.all(
+      this.getAllSubExchanges().map(async (subExchange) => {
+        await this._oracle.close();
```
xjcaa, line 364: close should be outside

src/exchange.ts
xjcaa: usdcMint

src/market.ts
```diff
@@ -191,47 +207,53 @@ export class ZetaGroupMarkets {
...
-   * Should not be called outside of Exchange.
+   * Should not be called outside of SubExchange.
    */
   public static async load(
+    asset: Asset,
     opts: ConfirmOptions,
     throttleMs: number
   ): Promise<ZetaGroupMarkets> {
-    let instance = new ZetaGroupMarkets();
+    let instance = new ZetaGroupMarkets(asset);
```
xjcaa, line 217: speaking for most things that are asset specific; 

could you just do

let subExchange = Exchange.getSubExchange(asset);

then use subExchange?

src/market.ts
```diff
@@ -242,27 +264,41 @@ export class ZetaGroupMarkets {
...
   /**
-   * Updates the option series state based off state in Exchange.
+   * Updates the option series state based off state in SubExchange.
    */
-  public async updateExpirySeries() {
-    for (var i = 0; i < Exchange.zetaGroup.products.length; i++) {
+  public async updateExpirySeries(asset: Asset) {
+    for (
+      var i = 0;
+      i < Exchange.getSubExchange(asset).zetaGroup.products.length;
```
xjcaa, line 277: also mentioned elsewhere; continually using `getSubExchange` seems unnecessarily expensive. 

Can just hold the ref to subexchange

src/market.ts
```diff
@@ -522,7 +579,7 @@ export class Market {
       side: order.side == "buy" ? Side.BID : Side.ASK,
       orderId: order.orderId,
       owner: order.openOrdersAddress,
-      clientOrderId: order.clientId,
+      clientOrderId: order.subClientId,
```
xjcaa, line 582: unintentional replace?

src/oracle.ts
```diff
@@ -47,61 +48,70 @@ export class Oracle {
...
-        },
-        Exchange.provider.connection.commitment
-      );
-      this._subscriptionIds.set(feed, subscriptionId);
 
-      // TODO set this so localnet has data for the oracle
-      // Remove once there is an oracle simulator.
-      await this.pollPrice(feed, false);
-    }
+    Promise.all(
```
xjcaa, line 81: await

src/risk.ts
```diff
@@ -12,61 +12,80 @@ import {
...
    * @param productIndex   market index of the product to calculate margin for.
    * @param size           signed integer of size for margin requirements (short orders should be negative)
    * @param marginType     type of margin calculation.
    */
   public getMarginRequirement(
+    asset: Asset,
     productIndex: number,
     size: number,
     marginType: MarginType
   ): number {
```
xjcaa, line 64: Not related to your PR, but should this return null instead of 0 or error ?

Seems misleading to return 0 margin requirement

src/subclient.ts
```diff
@@ -0,0 +1,2016 @@
...
+  MAX_SETTLE_AND_CLOSE_PER_TX,
+  MAX_CANCELS_PER_TX,
+  MAX_POSITION_MOVEMENTS,
+  DEFAULT_CLIENT_POLL_INTERVAL,
+  DEFAULT_CLIENT_TIMER_INTERVAL,
+  DEX_PID,
+  DEFAULT_ORDER_TAG,
+} from "./constants";
+import { exchange as Exchange } from "./exchange";
+
```
xjcaa, line 14: nit space

src/subclient.ts
```diff
@@ -0,0 +1,2016 @@
...
+  ProgramAccountType,
+  Wallet,
+  CancelArgs,
+  OrderType,
+  Position,
+  Order,
+  Side,
+  MovementType,
+} from "./types";
+import {
```
xjcaa, line 43: Side note: all these imports are huge - should we just

import * as instructions from "./program-instructions"

and access via instructions namespace etc

src/subclient.ts
```diff
@@ -0,0 +1,2016 @@
...
+  PositionMovementArg,
+} from "./program-instructions";
+
+import { EventType } from "./events";
+
+export class SubClient {
+  /**
+   * Returns the user wallet public key.
+   */
+  public get publicKey(): PublicKey {
```
xjcaa, line 73: global;

publickey, provider, program, usdcAccountAddress,

src/subclient.ts
```diff
@@ -0,0 +1,2016 @@
...
+   */
+  public static async load(
+    asset: Asset,
+    connection: Connection,
+    wallet: Wallet,
+    opts: ConfirmOptions = utils.defaultCommitment(),
+    callback: (asset: Asset, type: EventType, data: any) => void = undefined,
+    throttle: boolean = false
+  ): Promise<SubClient> {
+    console.log(
```
xjcaa, line 285: i wonder if worth storing a reference to the subexchange as part of client so you don't need to fetch it as that's constant.

src/subclient.ts
```diff
@@ -0,0 +1,2016 @@
...
+        subClient.updateSpreadPositions();
+
+        if (callback !== undefined) {
+          callback(asset, EventType.USER, null);
+        }
+      },
+      connection.commitment
+    );
+
+    subClient._usdcAccountAddress = await utils.getAssociatedTokenAddress(
```
xjcaa, line 352: btw side note: whether you want to change in this PR or another is fine w me;

they released findProgramAddressSync in one of the latest versions of solana web3, no more async functions for no reason

src/subclient.ts
```diff
@@ -0,0 +1,2016 @@
...
+  /**
+   * @param timerInterval   desired interval for subClient polling.
+   */
+  private setPolling(timerInterval: number) {
+    if (this._pollIntervalId !== undefined) {
+      console.log(`Resetting existing timer to ${timerInterval} seconds.`);
+      clearInterval(this._pollIntervalId);
+    }
+
+    this._pollIntervalId = setInterval(async () => {
```
xjcaa, line 434: As mentioned, it would be nice to move this interval to the parent, and just call into a function for the subclient to do the below logic.

src/utils.ts
```diff
@@ -50,6 +52,7 @@ import {
 } from "./program-instructions";
 import { Decimal } from "./decimal";
 import { readBigInt64LE } from "./oracle-utils";
+import { parseJsonText } from "typescript";
```
xjcaa, line 55: is this used

src/client.ts
filipzeta: idm, but having a map saves us in the future if assets don't end up counting up from 0 to N nicely - maybe if we have a bunch and then remove some? 

so imo better to use map and there's no real downside

src/client.ts
```diff
@@ -1,1033 +1,238 @@
...
-      );
-      openOrdersPda = _openOrdersPda;
-      tx.add(initIx);
-    } else {
-      openOrdersPda = this._openOrdersAccounts[marketIndex];
-    }
-
-    let orderIx = placeOrderIx(
-      marketIndex,
+    return await this.getSubClient(asset).placeOrderV3(
```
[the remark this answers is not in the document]
filipzeta: yep, intentionally made client.placeOrder() use the latest version

src/exchange.ts
filipzeta: i just use it in a few other places in exchange.ts, makes the code a bit more tidy. can remove if you want but imo doesnt take up too much space

src/exchange.ts
filipzeta: true, ill just remove updateAllAssetsOrderbooks. if we need functionality for all assets that doesnt exist rn, we'll find out during frontend dev and also zeta-options integration testing so ill add them in as needed.

src/risk.ts
filipzeta: yep agreed, misleading to return 0. ill return null.

src/utils.ts
```diff
@@ -91,14 +94,15 @@ export async function getSettlement(
 }
```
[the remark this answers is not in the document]
filipzeta: added another ticket: https://www.notion.so/sierraresearch/SDK-getXYZaddress-rename-5982b620044941f79082daec8d75adf4

better to do it separately so that it's easier to run all the tests and say yep nothing broke

src/subclient.ts
filipzeta: also made a ticket for this one :) https://www.notion.so/sierraresearch/Migrate-to-findProgramAddressSync-in-SDK-b77128b690a34d2a9b3a0396b8b20727

src/subclient.ts
filipzeta: yes, i also changed a few others - errors, types, constants. they were all inconsistent across different files 😝

src/subclient.ts
filipzeta: good idea, makes the code a lot more clean

src/subclient.ts
filipzeta: done - does the way ive split it look OK? we keep lastUpdateTimestamp, pollInterval and pendingUpdate inside `subClient`

examples/basic/basic-example.ts
```diff
@@ -7,106 +7,211 @@ import {
...
+  events,
 } from "@zetamarkets/sdk";
-import { PublicKey, Connection, Keypair } from "@solana/web3.js";
+import { PublicKey, Connection, Keypair, SOLANA_SCHEMA } from "@solana/web3.js";
 import fetch from "node-fetch";
+import { airdropUsdc } from "../liquidator-example/src/utils";
+import { Token, TOKEN_PROGRAM_ID } from "@solana/spl-token";
+import {} from "../../dist/subexchange";
+import { defaultCommitment } from "../../dist/utils";
+// import { Exchange } from "../../src/exchange";
```
xjcaa, line 19: comments + could you import from the sdk rather than dist/

examples/basic/basic-example.ts
```diff
@@ -7,106 +7,211 @@ import {
   Network,
   utils,
   types,
+  assets,
+  events,
 } from "@zetamarkets/sdk";
-import { PublicKey, Connection, Keypair } from "@solana/web3.js";
+import { PublicKey, Connection, Keypair, SOLANA_SCHEMA } from "@solana/web3.js";
```
xjcaa, line 13: what's SOLANA_SCHEMA used for

README.md
```diff
@@ -334,9 +334,9 @@ await client.placeOrder(
 
 // View our position
 await client.updateState();
-console.log(client.positions);
+console.log(client.marginPositions);
 
-// `client.positions` is a list of positions in market index order.
+// `client.marginPositions` is a list of marginAccount positions in market index order.
```
xjcaa, line 339: Add this to breaking changes

src/client.ts
```diff
@@ -384,30 +134,46 @@ export class Client {
...
+      })
+    );
+
+    client.setPolling(constants.DEFAULT_CLIENT_TIMER_INTERVAL);
 
-    client.setPolling(DEFAULT_CLIENT_TIMER_INTERVAL);
     return client;
   }
 
+  public async addSubClient(asset: Asset, subClient: SubClient) {
```
xjcaa, line 163: this doesn't have to be async

src/client.ts
```diff
@@ -416,1471 +182,421 @@ export class Client {
...
+    );
+  }
 
-      try {
-        this._spreadAccount = (await this._program.account.spreadAccount.fetch(
-          this._spreadAccountAddress
-        )) as unknown as SpreadAccount;
-      } catch (e) {}
-    }
+  public async deposit(
```
xjcaa, line 214: feature request, could we add an instruction that's called migrateBalance which is atomic withdrawal from one, deposit to another

src/client.ts
```diff
@@ -416,1471 +182,421 @@ export class Client {
...
+    } catch (e) {
+      throw Error(
+        "User has no USDC associated token account. Please create one and deposit USDC."
+      );
     }
+  }
 
-    if (this._spreadAccount !== null) {
-      this.updateSpreadPositions();
+  public async updateState(
```
xjcaa, line 240: Should add docs to what these args mean

src/client.ts
```diff
@@ -416,1471 +182,421 @@ export class Client {
...
-   * to cancel the second order with the same client order id,
-   * you may need to crank the corresponding event queue to flush that order id
-   * from the user open orders account before cancelling the second order.
-   * (Depending on the order in which the order was cancelled).
-   */
-  public async placeOrder(
-    market: PublicKey,
+  public async placeOrderAndLockPosition(
+    asset: Asset,
+    marketIndex: number,
```
xjcaa, line 320: Changing placeOrder to use marketIndex instead of market publicKey is somewhat a breaking change, how about adding a type that is

something like

```
export type MarketAddress = PublicKey;
export type MarketIndex = number;
export type MarketIndentifier = MarketAddress | MarketIndex;

OR

export type MarketIndentifier = number | PublicKey;
```

xjcaa: And we can have this consistent across subclient place orders as well

src/client.ts
```diff
@@ -1,121 +1,74 @@
...
+    return positionMap;
   }
-  private _marginAccount: MarginAccount | null;
 
-  /**
-   * Client margin account address.
-   */
-  public get marginAccountAddress(): PublicKey {
-    return this._marginAccountAddress;
+  public get spreadPositions(): Map<Asset, types.Position[]> {
```
xjcaa, line 47: Do you reckon it's better to make people access their positions directly through the sub client. or

client.getPositions(asset: Asset)

xjcaa: Or, what if we modify `types.Position` to hold asset? then positions is just generically Position[]

src/exchange.ts
```diff
@@ -167,7 +158,7 @@ export class Exchange {
   public get clockTimestamp(): number {
     return this._clockTimestamp;
   }
-  private _clockTimestamp: number;
+  private _clockTimestamp: number = 0;
```
xjcaa, line 161: can default to undefined

src/exchange.ts
```diff
@@ -215,74 +189,61 @@ export class Exchange {
...
-  private init(
+  public async initialize(
+    assets: Asset[],
     programId: PublicKey,
     network: Network,
     connection: Connection,
-    wallet: Wallet,
-    opts?: ConfirmOptions
+    wallet = new types.DummyWallet(),
+    opts: ConfirmOptions
```
xjcaa, line 214: I think semantically, fields with default options should be at the back, which is quite practical for Exchange load, otherwise, we can set default value for opts as well which as available via `utils.defaultCommitment`

src/exchange.ts
```diff
@@ -712,441 +447,229 @@ expirationThresholdSeconds=${params.expirationThresholdSeconds}`
...
   }
 
   /**
-   * Update pricing for an expiry index.
+   * Helper function to get the deposit limits
    */
-  public async updatePricing(expiryIndex: number) {
-    let tx = new Transaction().add(instructions.updatePricingIx(expiryIndex));
-    await utils.processTransaction(this._provider, tx);
+  public async getDepositLimit(): Promise<number> {
```
xjcaa, line 463: not sure if we need this tbh

src/exchange.ts
```diff
@@ -712,441 +447,229 @@ expirationThresholdSeconds=${params.expirationThresholdSeconds}`
...
+    asset: Asset,
+    args: instructions.UpdateInterestRateArgs
+  ) {
+    await this.getSubExchange(asset).updateInterestRate(args);
   }
 
-  public async updateInterestRate(args: instructions.UpdateInterestRateArgs) {
-    let tx = new Transaction().add(
-      instructions.updateInterestRateIx(args, this._provider.wallet.publicKey)
+  public async close() {
```
xjcaa, line 659: I think you'd unsubscribe from the program subscription ids in this one since `addProgramSubscriptionId` is in this one.

src/market.ts
```diff
@@ -319,19 +305,22 @@ export class ZetaGroupMarkets {
 }
 
 export class ExpirySeries {
```
xjcaa, line 307: Does this need asset, since it belongs to a ZetaGroupMarket ?

src/market.ts
```diff
@@ -367,17 +356,27 @@ export class Market {
     return this._expiryIndex;
   }
   public get expirySeries(): ExpirySeries {
```
xjcaa, line 358: I realise this existed before, but thinking about it this function doesn't make sense? shouldn't it just return itself? LOL

src/oracle.ts
```diff
@@ -62,46 +64,54 @@ export class Oracle {
...
+              lastUpdatedSlot: priceData.publishSlot,
+            };
+            this._data.set(feed, oracleData);
+            this._callback(asset, oracleData);
+          },
+          Exchange.provider.connection.commitment
+        );
+
+        this._subscriptionIds.set(feed, subscriptionId);
+        // TODO set this so localnet has data for the oracle
```
xjcaa, line 110: This comment can just be:

`Set this so the oracle contains a price on initialization.`

src/oracle.ts
```diff
@@ -3,13 +3,14 @@ import { parsePythData, Price } from "./oracle-utils";
 import { Network } from "./network";
```
xjcaa, line 3: What are your thoughts on deprecating oracle feed i.e. "SOL/USD" and just make the feeds of type Asset.

src/risk.ts
```diff
@@ -1,72 +1,85 @@
...
-      return;
+  public updateMarginRequirements(asset: Asset) {
+    if (
+      Exchange.getSubExchange(asset).greeks === undefined ||
+      Exchange.oracle === undefined
+    ) {
+      throw Error("Pricing (greeks and/or oracle) is not initialized");
     }
-    let oraclePrice = Exchange.oracle.getPrice("SOL/USD");
+    let oraclePrice = Exchange.oracle.getPrice(assetToOracleFeed(asset));
```
xjcaa, line 35: Motivation behind the oracle comment -> Seems convoluted :p 

Should just be getPrice(asset)?

src/subclient.ts
```diff
@@ -0,0 +1,1822 @@
...
+  public static async load(
+    asset: Asset,
+    parentClient: Client,
+    connection: Connection,
+    wallet: types.Wallet,
+    callback: (asset: Asset, type: EventType, data: any) => void = undefined,
+    throttle: boolean = false
+  ): Promise<SubClient> {
+    let subClient = new SubClient(asset);
+    subClient._parentClient = parentClient;
```
xjcaa, line 202: you could put this in the constructor

src/subclient.ts
```diff
@@ -0,0 +1,1822 @@
...
+  }
+  private _asset: Asset;
+
+  /**
+   * A reference to this subclient's parent, to grab things like usdc and whitelist addresses
+   */
+  public get parentClient(): Client {
+    return this._parentClient;
+  }
+  private _parentClient: Client;
```
xjcaa, line 58: could just call it parent if you wanted :p

src/subclient.ts
```diff
@@ -0,0 +1,1822 @@
...
+   * @param size            the quantity of the order (3 d.p)
+   * @param side            the side of the order. bid / ask
+   * @param clientOrderId   optional: subClient order id (non 0 value)
+   * NOTE: If duplicate subClient order ids are used, after a cancel order,
+   * to cancel the second order with the same subClient order id,
+   * you may need to crank the corresponding event queue to flush that order id
+   * from the user open orders account before cancelling the second order.
+   * (Depending on the order in which the order was cancelled).
+   */
+  public async placeOrder(
```
xjcaa, line 507: Should we deprecate v1/v2?

src/subclient.ts
```diff
@@ -0,0 +1,1822 @@
...
+   * you may need to crank the corresponding event queue to flush that order id
+   * from the user open orders account before cancelling the second order.
+   * (Depending on the order in which the order was cancelled).
+   */
+  public async placeOrderV3(
+    market: PublicKey,
+    price: number,
+    size: number,
+    side: types.Side,
+    orderType: types.OrderType,
```
xjcaa, line 731: Let's default this to limit if we deprecate v1/v2

src/subexchange.ts
```diff
@@ -0,0 +1,870 @@
...
+        Exchange.programId,
+        this.zetaGroupAddress
+      );
+
+    const [socializedLossAccount, _socializedLossAccountNonce] =
+      await utils.getSocializedLossAccount(
+        Exchange.programId,
+        this._zetaGroupAddress
+      );
+    this;
```
xjcaa, line 172: ?

src/subexchange.ts
```diff
@@ -0,0 +1,870 @@
...
+  }
+
+  public addProgramSubscriptionId(id: number) {
+    this._programSubscriptionIds.push(id);
+  }
+
+  /**
+   * Close the websockets.
+   */
+  public async close() {
```
xjcaa, line 668: Move to bottom

src/subexchange.ts
```diff
@@ -0,0 +1,870 @@
...
+          let txSig = await utils.processTransaction(Exchange.provider, tx);
+          console.log(`[REBALANCE INSURANCE VAULT]: ${txSig}`);
+        })
+      );
+    } catch (e) {
+      console.log(`Error in rebalancing the insurance vault ${e}`);
+    }
+  }
+
+  public addProgramSubscriptionId(id: number) {
```
xjcaa, line 661: chuck this into the parent

src/utils.ts
```diff
@@ -14,42 +14,27 @@ import {
...
-  settlePositionsIx,
-  settleSpreadPositionsIx,
-  PositionMovementArg,
-} from "./program-instructions";
+import { Network } from "./network";
+import * as types from "./types";
+import * as instructions from "./program-instructions";
 import { Decimal } from "./decimal";
 import { readBigInt64LE } from "./oracle-utils";
+import { decode } from "punycode";
```
xjcaa, line 36: wat this

examples/basic/basic-example.ts
filipzeta: unused, removed

filipzeta:
- [ ] Clean up all imports before merging

src/exchange.ts
filipzeta: `addProgramSubscriptionId` in `Exchange` just calls the corresponding `SubExchange` function, so it doesn't feel right to remove the _programSubscriptionIds in `Exchange` because that's part of `close()` in `SubExchange`

src/market.ts
filipzeta: Technically not necessary to have it but it's just handy rather than passing `asset` every time you call a function

src/utils.ts
filipzeta: ah probs automatically done, removed

filipzeta:
- [ ] Reformat how greeks are provided by the SDK - @wheeeliam gave some examples, we calculate the productGreekIndex in a pretty unintuitive way

src/utils.ts
```diff
@@ -1123,8 +1163,37 @@ export async function writeKeypair(filename: string, keypair: Keypair) {
 }
 
 export async function getAllProgramAccountAddresses(
-  accountType: ProgramAccountType
+  accountType: types.ProgramAccountType,
+  asset: assets.Asset | undefined
```
xjcaa, line 1167: `asset: assets.Asset = undefined` ?

src/utils.ts
```diff
@@ -1123,8 +1163,37 @@ export async function writeKeypair(filename: string, keypair: Keypair) {
...
+          anchor.BorshAccountsCoder.accountDiscriminator(accountType)
+        ),
+      },
+    },
+  ];
+
+  if (asset != undefined) {
+    let assetOffset = 0;
+    // From the account itself in account.rs
+    if (accountType == types.ProgramAccountType.MarginAccount) {
```
xjcaa, line 1183: have you tested this for both spread n margin?

src/utils.ts
```diff
@@ -1303,3 +1385,39 @@ export function calculateMovementFees(
...
+    );
+  } else {
+    // File does not exist
+    keypair = Keypair.generate();
+    writeKeypair(filename, keypair);
+  }
+  return keypair;
+}
+
+export function toNetwork(network: string): Network {
```
xjcaa, line 1410: add to network.ts?

src/client.ts
filipzeta: Ideally I don't want anyone digging through SubClient, it should all go through Client. I think having client.getMarginPositions(asset) is fine, but it's pretty much the same thing as what I have now which would be client.marginPositions().get(asset)

filipzeta: Actually I change my mind, I think client.getMarginPositions(asset) is better because it's consistent with how we do orders which is client.getOrders(asset).

If there's a need to grab all at once in the future we can add an additional function for all assets, or users can just call it in a Promise.all

src/exchange.ts
filipzeta: nothing, confirmed with FE and removed

src/risk.ts
```diff
@@ -1,72 +1,85 @@
...
 } from "./utils";
+import { assetToOracleFeed, Asset } from "./assets";
 
 export class RiskCalculator {
   /**
    * Returns the margin requirements for all markets,
    * indexed by market index.
    */
-  public get marginRequirement(): Array<MarginRequirement> {
+  public get marginRequirement(): Map<Asset, Array<types.MarginRequirement>> {
```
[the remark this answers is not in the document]
filipzeta: yes thats a good idea, ive removed it in favour of a getMarginRequirements(asset) function

src/subclient.ts
filipzeta: Yeah the idea here is a soft removal of v1/v2, users can't use it unless they're on old SDK or manually calling the subClient, and then when people have moved over we can delete v1/v2.

Should we just delete them now instead?

src/utils.ts
filipzeta: yes with the indexers

filipzeta:
- [ ] Have some sort of error if we try to access Exchange.getSubExchange() before loading

src/market.ts
filipzeta: im a bit confused 😅 its calling `markets.expirySeries[index]` right now isn't it? and it seems like it has to do the whole `Exchange.` stuff as there's no way to go back up to `.markets` from where we are in `.markets.markets`

src/oracle.ts
filipzeta: yep agreed that's a good idea!

src/client.ts
filipzeta: added and tested locally
image

src/exchange.ts
```diff
@@ -361,6 +365,14 @@ export class Exchange {
       })
     );
 
+    await Promise.all(
+      this.assets.map(async (asset) => {
+        this.getMarkets(asset).map(async (market) => {
```
xjcaa, line 370: doesn't need to be async, and i think you can just copy from this.getMarkets?

src/exchange.ts
```diff
@@ -483,10 +495,73 @@ export class Exchange {
...
+
+  public unsubscribeMarket(asset: Asset, index: number) {
+    this.getSubExchange(asset).markets.unsubscribeMarket(index);
+  }
+
   public async updateOrderbook(asset: Asset, index: number) {
     await this.getSubExchange(asset).markets.markets[index].updateOrderbook();
   }
 
+  public async updateAllLiveOrderbooks() {
```
xjcaa, line 510: change to `updateAllOrderbooks(live = true)`? and let people decide whether to filter

filipzeta:
- [ ] Make `fromProgramAsset` nicer after all merges and merge conflicts

CHANGELOG.md
```diff
@@ -3,7 +3,9 @@
 All notable changes to this project will be documented in this file.
 Version changes are pinned to SDK releases.
 
-## [Unreleased]
+## Unreleased
+
+## [0.15.0]
```
xjcaa, line 9: When you release, just make a note that we've skipped 15.0 due to versioning issues.

src/assets.ts
```diff
@@ -10,10 +11,39 @@ export enum Asset {
...
+    }
+  }
+  return allAssets;
+}
+
 export function assetToName(asset: Asset) {
   if (asset == Asset.SOL) return "SOL";
   if (asset == Asset.BTC) return "BTC";
   if (asset == Asset.ETH) return "ETH";
+  if (asset == null) return ""; // Some things, like clock callbacks, are for all assets and return asset=null
```
xjcaa, line 46: Should this function just return string | null? and return null on that case.

src/client.ts
```diff
@@ -224,170 +76,47 @@ export class Client {
...
-
     client._whitelistDepositAddress = undefined;
     try {
       let [whitelistDepositAddress, _whitelistTradingFeesNonce] =
         await utils.getUserWhitelistDepositAccount(
           Exchange.programId,
           wallet.publicKey
         );
-      await client._program.account.whitelistDepositAccount.fetch(
+      await Exchange.program.account.whitelistDepositAccount.fetch(
```
xjcaa, line 119: Yeah nice makes sense to deprecate client.program.

src/client.ts
```diff
@@ -475,1498 +220,494 @@ export class Client {
...
     }, timerInterval * 1000);
   }
 
-  private toggleUpdateState(toggleOn: boolean) {
-    if (toggleOn) {
-      this._updatingState = true;
-      this._updatingStateTimestamp = Date.now() / 1000;
+  public marketIdentifierToPublicKey(
+    asset: Asset,
+    marketIndex: types.MarketIdentifier
```
xjcaa, line 233: this name shouldn't be marketIndex if it's also a pubkey

src/client.ts
```diff
@@ -475,1498 +220,494 @@ export class Client {
...
+        depositSubClient.marginAccountAddress,
+        this.usdcAccountAddress,
         this.publicKey,
-        this._whitelistDepositAddress
+        this.whitelistDepositAddress
       )
     );
+
+    // Send
     let txId = await utils.processTransaction(this._provider, tx);
```
xjcaa, line 311: can return directly here

src/client.ts
xjcaa: `market` seems better?

src/constants.ts
```diff
@@ -1,11 +1,13 @@
 import { PublicKey } from "@solana/web3.js";
+import { Asset } from "./assets";
 
+// TODO add comment;
```
xjcaa, line 4: delete todo?

src/constants.ts
```diff
@@ -1,11 +1,13 @@
...
+// Asset keys are wormhole from mainnet.
 export const MINTS = {
-  SOL: new PublicKey("So11111111111111111111111111111111111111112"),
+  [Asset.SOL]: new PublicKey("So11111111111111111111111111111111111111112"),
+  [Asset.BTC]: new PublicKey("qfnqNqs3nCAHjnyCgLRDbBtq4p2MtHZxw8YjSyYhPoL"),
```
xjcaa, line 8: Let's add ETH for completedness

src/market.ts
xjcaa: oh right you are correct

src/oracle.ts
```diff
@@ -3,13 +3,15 @@ import { parsePythData, Price } from "./oracle-utils";
...
 import { exchange as Exchange } from "./exchange";
 import * as constants from "./constants";
+import { Asset, assetToName } from "./assets";
+import { assets } from ".";
```
xjcaa, line 7: nit "./"

src/oracle.ts
```diff
@@ -43,65 +45,71 @@ export class Oracle {
...
   // Fetch and update an oracle price manually
   public async pollPrice(
-    feed: string,
+    asset: assets.Asset,
     triggerCallback = true
   ): Promise<OraclePrice> {
-    if (!(feed in constants.PYTH_PRICE_FEEDS[this._network])) {
-      throw Error("Invalid Oracle feed!");
+    if (!(asset in constants.PYTH_PRICE_FEEDS[this._network])) {
+      throw Error("Invalid Oracle feed, no matching pubkey!");
```
xjcaa, line 52: No matching asset?

src/subclient.ts
xjcaa: yeah let's just only have v3 in the SDK

src/types.ts
```diff
@@ -176,6 +176,14 @@ export interface CancelArgs {
   cancelSide: Side;
 }
 
+export interface InitializeMarginAccountArgs {
+  asset: Asset;
+}
+
+export interface InitializeSpreadAccountArgs {
```
xjcaa, line 183: what are these used for again

src/assets.ts
filipzeta: yep, also added explicit return types to other functions

src/client.ts
filipzeta: yeah `market` is better, ill rename the occurrences

src/oracle.ts
```diff
@@ -3,13 +3,15 @@ import { parsePythData, Price } from "./oracle-utils";
...
 
 export class Oracle {
   private _connection: Connection;
   private _network: Network;
-  private _data: Map<string, OraclePrice>;
-  private _subscriptionIds: Map<string, number>;
-  private _callback: (price: OraclePrice) => void;
+  private _data: Map<assets.Asset, OraclePrice>;
+  private _subscriptionIds: Map<assets.Asset, number>;
+  private _callback: (asset: Asset, price: OraclePrice) => void;
```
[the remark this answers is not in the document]
filipzeta: yeah looks a bit weird aye, ill leave it as `import { assets } from "./";` only

src/types.ts
filipzeta: T_T i didnt remove them, they are remnants of an older time haha

src/utils.ts
```diff
@@ -1146,8 +1189,37 @@ export async function writeKeypair(filename: string, keypair: Keypair) {
...
+        ),
+      },
+    },
+  ];
+
+  if (asset != undefined) {
+    let assetOffset = 0;
+    // From the account itself in account.rs
+    if (accountType == types.ProgramAccountType.MarginAccount) {
+      assetOffset = 8 + 32 + 1 + 8 + 1 + 138 + 48 + 5520 + 8;
```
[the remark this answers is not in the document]
filipzeta: yep, ill add the individual bits as a comment too

xjcaa (approved):
send

PR #126, opened 2022-06-08
Add asset to marginaccount and spreadaccount

filipzeta:
For multi-underlying support we want to know the asset of each margin and spread account. This adds it, and also forces any new accounts from here to pass in the asset.

### DIFF
--- a/src/program-types.ts
+++ b/src/program-types.ts
@@ -96,6 +96,7 @@ export interface ZetaGroup {
   expirySeries: Array<ExpirySeries>;
   expirySeriesPadding: Array<ExpirySeries>;
   totalInsuranceVaultDeposits: anchor.BN;
+  asset: any;
   padding: Array<number>;
 }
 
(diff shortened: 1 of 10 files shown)


src/program-types.ts
```diff
@@ -146,6 +146,7 @@ export interface MarginAccount {
   productLedgers: Array<ProductLedger>;
   productLedgersPadding: Array<ProductLedger>;
   rebalanceAmount: anchor.BN;
+  asset: any;
```
filipzeta, line 149: is this the best way to handle the type? if i put `asset: number` i get 

```
Types of property 'asset' are incompatible.
    Type 'TypeDef<{ name: "Asset"; type: { kind: "enum"; variants: [{ name: "SOL"; }, { name: "BTC"; }]; }; } & { name: "Asset"; }, Record<string, never>>' is not comparable to type 'number'
```

when i try to read it over in `utils.ts` (in `getAllOpenOrdersAccountsByMarket()` we do a `let marginAccount = acc.account as MarginAccount;`)

filipzeta:
- [ ] Add a bunch of other underlyings to assets.ts

src/program-types.ts
xjcaa: Correct it is

xjcaa (approved):

PR #135, opened 2022-07-21
Add OrderCompleteEvent

filipzeta:
Mostly autogenerated PR from https://github.com/zetamarkets/zeta-options/pull/453/. Bigger description is available over in that PR :)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@ Version changes are pinned to SDK releases.
 
 ## Unreleased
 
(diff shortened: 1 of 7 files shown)


xjcaa:
Can we add a subscription to the margin account for order complete events and add a new callback

filipzeta:
> Can we add a subscription to the margin account for order complete events and add a new callback

Added :)

kelvinlau20100 (approved):
Yeah just the order booted event as well and we should be good to go

xjcaa (approved):

PR #136, opened 2022-07-22
Fee-collection-rework

kelvinlau20100:
collectTreasuryFunds isn't exposed similar to updateAdmin.


src/exchange.ts
```diff
@@ -251,12 +259,17 @@ export class Exchange {
     const [serumAuthority, serumNonce] = await utils.getSerumAuthority(
       this.programId
     );
+    const [treasuryWallet, _treasuryWalletnonce] =
```
filipzeta, line 262: Should this be in Exchange.load() instead? initializeZetaState() and initializeZetaGroup() are only used in tests and when we want to make new ZetaState or ZetaGroup accounts - in real world use we'd just call Exchange.load() and that's it

src/program-instructions.ts
```diff
@@ -912,16 +912,56 @@ export async function initializeZetaGroupIx(
...
+  amount: anchor.BN
+): TransactionInstruction {
+  let subExchange = Exchange.getSubExchange(asset);
+  return Exchange.program.instruction.treasuryMovement(
+    types.toProgramMovementType(movementType),
+    amount,
+    {
+      accounts: {
+        state: Exchange.stateAddress,
+        zetaGroup: subExchange.zetaGroupAddress,
```
filipzeta, line 943: nit: can just do Exchange.getZetaGroupAddress(asset) and similar for others

src/exchange.ts
```diff
@@ -272,7 +286,7 @@ export class Exchange {
     this._mintAuthority = mintAuthority;
     this._stateAddress = state;
     this._serumAuthority = serumAuthority;
-    this._usdcMintAddress = constants.USDC_MINT_ADDRESS[this.network];
```
xjcaa, line 275: why is this removed

xjcaa (approved):

PR #140, opened 2022-08-03
Referral Rewards

kelvinlau20100:


src/client.ts
```diff
@@ -788,6 +794,36 @@ export class Client {
...
+    let tx = new Transaction().add(
+      await instructions.claimReferralsRewardsIx(
+        referralAccountAddress,
+        this._usdcAccountAddress,
+        this.publicKey
+      )
+    );
+    return await utils.processTransaction(this._provider, tx);
+  }
+
```
kelvinlau20100, line 826: could leave as is or combine into a single function

src/exchange.ts
```diff
@@ -251,7 +259,10 @@ export class Exchange {
     this._isSetup = true;
   }
 
-  public async initializeZetaState(params: instructions.StateParams) {
+  public async initializeZetaState(
+    params: instructions.StateParams,
+    referralAdmin: PublicKey
```
kelvinlau20100, line 264: could include this as part of StateParams but i feel like it's slightly different

filipzeta (approved):
chee hoo

PR #141, opened 2022-08-08
Clock polling boolean

filipzeta:
We have seen a few instances where prolonged RPC issues lead to a webserver outage. One problem that I've found is that we poll multiple accounts on a timer (`DEFAULT_EXCHANGE_POLL_INTERVAL`), but if one of them does not succeed then another timer interval is begun on top of that. For example:
image
Here we see 1 `caught exception in orderbook poll function`, which has a similar boolean mechanism, and N `SubExchange polling failed`, which we're fixing here.  When the RPC issues are resolved we now have all of those N timers happening at once and the application grinds to a halt and needs a restart.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ Version changes are pinned to SDK releases.
 
 ## [Unreleased]
 
+## [0.16.5]
+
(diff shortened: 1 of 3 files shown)


filipzeta:
PR not necessary, `this._lastPollTimestamp` is already performing the same functionality

PR #144, opened 2022-08-11
Add calculate position movement function to risk calculator

wheeeliam:


src/risk.ts
```diff
@@ -533,3 +651,470 @@ export function calculateLongOptionMargin(
...
+    if (costOfTrades !== 0)
+      throw Error("Cost of trades must be greater than zero.");
+    return;
+  }
+  let ledger = marginAccount.productLedgers[index];
+  let [openSize, closeSize] = getExecutionOpenCloseSize(
+    ledger.position.size.toNumber(),
+    size
+  );
+  let sideIndex = size > 0 ? 0 : 1;
```
filipzeta, line 864: nit: constants.BID_ORDERS_INDEX/ASK_ORDERS_INDEX?

src/risk.ts
```diff
@@ -533,3 +651,470 @@ export function calculateLongOptionMargin(
...
+    ledger.position.size.toNumber() > 0
+      ? constants.ASK_ORDERS_INDEX
+      : constants.BID_ORDERS_INDEX;
+  ledger.orderState.closingOrders = new BN(
+    Math.min(
+      Math.abs(ledger.position.size.toNumber()),
+      ledger.orderState.openingOrders[index].toNumber()
+    )
+  );
+  ledger.orderState.openingOrders[index] = new BN(
```
filipzeta, line 921: nit: I think there's some BN -> number -> BN going on here and above that can be simplified

src/risk.ts
```diff
@@ -533,3 +651,470 @@ export function calculateLongOptionMargin(
...
+}
+
+/// Moves closing orders to the respective opening orders.
+/// To be called before executing or moving positions.
+/// `rebalance_orders` should be called after executing.
+function resetClosingOrders(ledger: ProductLedger) {
+  if (
+    ledger.orderState.closingOrders.toNumber() === 0 ||
+    ledger.position.size.toNumber() === 0
+  )
```
filipzeta, line 934: very nitpicky but i personally don't like the single-line `if` syntax where there isn't a curly bracket, just because when you go to edit it later you can get sneaky logic errors if you don't remember to add curly brackets. 

you've got some inconsistent usage of it, eg line 955 and line 1006 has curly brackets but here and 856 you don't

src/risk.ts
```diff
@@ -533,3 +651,470 @@ export function calculateLongOptionMargin(
...
+      ledger.orderState.closingOrders.toNumber()
+  );
+  ledger.orderState.closingOrders = new BN(0);
+}
+
+function moveSize(position: Position, size: number) {
+  if (size === 0) return 0;
+  if (
+    (size > 0 && position.size.toNumber() < size) ||
+    (size < 0 && position.size.toNumber() > size)
```
filipzeta, line 953: reminds me of this lol
image

src/risk.ts
```diff
@@ -533,3 +651,470 @@ export function calculateLongOptionMargin(
...
+    );
+  }
+}
+
+function prorataCostOfTrades(position: Position, size: number) {
+  let sizeAbs = Math.abs(position.size.toNumber());
+  if (size === sizeAbs) {
+    return position.costOfTrades.toNumber();
+  } else {
+    // TODO: Double check division of BN -> numbers
```
filipzeta, line 1109: Did you double check this? 😄

filipzeta (approved):
Overall looks good, the only comments I have aren't vital to the functionality :) Sucks that you had to essentially rewrite a bunch of rust code, but good that you formatted it very similarly and kept function names similar, made the review easy.

I compared side-by-side with the equivalent rust code and it all matches, but didn't worry about the actual logic itself as long as it matched because it would've already been tested when it was written.

filipzeta:
You'll need to:
- [ ] Bump the SDK version - edit package.json
- [ ] Add your changes to the changelog
- [ ] Release the SDK (npm publish)

Merge this PR before your zeta-options PR

src/risk.ts
```diff
@@ -533,3 +651,470 @@ export function calculateLongOptionMargin(
...
+    }
+  }
+
+  let spreadMarginRequirements = calculateSpreadAccountMarginRequirement(
+    spreadAccount,
+    zetaGroup
+  );
+
+  if (spreadMarginRequirements > spreadAccount.balance.toNumber()) {
+    let diff = spreadMarginRequirements - spreadAccount.balance.toNumber();
```
kelvinlau20100, line 821: if possible can we do BN arithmetic here instead

src/risk.ts
```diff
@@ -533,3 +651,470 @@ export function calculateLongOptionMargin(
...
+      throw Error("Insufficient funds to collateralize spread account.");
+    }
+    // Move funds from margin to spread.
+    spreadAccount.balance = new BN(spreadAccount.balance.toNumber() + diff);
+    marginAccount.balance = new BN(marginAccount.balance.toNumber() - diff);
+  } else if (spreadMarginRequirements < spreadAccount.balance.toNumber()) {
+    let diff = spreadAccount.balance.toNumber() - spreadMarginRequirements;
+    // Move funds from spread to margin.
+    spreadAccount.balance = new BN(spreadAccount.balance.toNumber() - diff);
+    marginAccount.balance = new BN(marginAccount.balance.toNumber() + diff);
```
kelvinlau20100, line 832: all calculations similar to this as well

src/risk.ts
```diff
@@ -533,3 +651,470 @@ export function calculateLongOptionMargin(
...
+  if (size === sizeAbs) {
+    return position.costOfTrades.toNumber();
+  } else {
+    // TODO: Double check division of BN -> numbers
+    return Math.floor(position.costOfTrades.toNumber() / sizeAbs) * size;
+  }
+}
+
+function calculateSignedCostOfTrades(price: number, size: number): number {
+  return (price * size) / Math.pow(10, constants.POSITION_PRECISION);
```
kelvinlau20100, line 1115: here we would also want to be careful when dividing by Math.pow(10, constants.POSITION_PRECISION)
in cases where price * size < 1000, you will get a decimal number because you're returning a number when we should be truncating it to 0 due to underflow so i would still Math,floor this as well, 

^ This is in the case that price is a native number and size is nativeLotSize number too

src/risk.ts
```diff
@@ -533,3 +651,470 @@ export function calculateLongOptionMargin(
...
+}
+
+function calculateNormalizedCostOfTrades(price: number, size: number): number {
```
kelvinlau20100, line 1118: this too

kelvinlau20100 (commented):
Not a big deal but where things like account.balance are BN
i would prefer to turn the number you're adding to it into a BN and then add that to an existing BN rather than BN to number and then numbr + number or number - number and then back into a BN, as that causes more chances for overflow to occur in code.

Another thing is i would pref any non risk calculator functions to be put into a risk-utils.ts file please. (a nit but my personal preference)

src/risk.ts
```diff
@@ -315,6 +326,113 @@ export class RiskCalculator {
...
+        asset: assets.toProgramAsset(asset),
+        padding: new Array(262).fill(0),
+      };
+    }
+
+    // Copy account objects to perform simulated position movement
+    let simulatedMarginAccount = cloneDeep(marginAccount);
+    let simulatedSpreadAccount = cloneDeep(spreadAccount);
+
+    let nativeSpot = utils.convertDecimalToNativeInteger(
```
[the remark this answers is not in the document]
wheeeliam: Works better - nice one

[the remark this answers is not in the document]
wheeeliam: This also means that the function doesn't need to be async either :)

src/risk-utils.ts
```diff
@@ -0,0 +1,713 @@
...
+
+  let spreadMarginRequirements = calculateSpreadAccountMarginRequirement(
+    spreadAccount,
+    zetaGroup
+  );
+
+  if (spreadMarginRequirements > spreadAccount.balance.toNumber()) {
+    let diff = spreadMarginRequirements - spreadAccount.balance.toNumber();
+    if (diff > marginAccount.balance.toNumber()) {
+      throw Error("Insufficient funds to collateralize spread account.");
```
xjcaa, line 406: Should this function error like this? and wouldn't you want more meaningful data i.e. you're missing (balance - diff) in your margin account

src/risk.ts
```diff
@@ -315,221 +324,117 @@ export class RiskCalculator {
...
-  let spotPercentageLong =
-    marginType == types.MarginType.INITIAL
-      ? subExchange.marginParams.optionSpotPercentageLongInitial
-      : subExchange.marginParams.optionSpotPercentageLongMaintenance;
-
-  return Math.min(
-    markPrice * markPercentageLong,
-    spotPrice * spotPercentageLong
-  );
+    return {
```
xjcaa, line 433: My only thought is since this is off chain, you can enrich all of this with more information on like why the movement fails etc... 

i.e. you can have invalid movements simulated and show a negative balance say, that has invalid state, but just show it as an error

src/risk-utils.ts
wheeeliam:
```    let spread_margin_requirements = spread_account.calculate_margin_requirement(&zeta_group)?;
    if spread_margin_requirements > spread_account.balance {
        let diff = spread_margin_requirements
            .checked_sub(spread_account.balance)
            .unwrap();
        if diff > margin_account.balance {
            return Err(error!(
                ZetaError::InsufficientFundsToCollateralizeSpreadAccount
            ));
        }```

wheeeliam: Agree with showing meaningful data - can add that in

PR #145, opened 2022-08-12
adding IOC tx

epsilonsized:
Adding method to send an order and cancel it without error as an IOC.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -555,21 +555,19 @@ export class Client {
   public async immediateOrCancel(
     asset: Asset,
     market: PublicKey,
-    newOrderClientOrderId: number,
-    newOrderPrice: number,
-    newOrderSize: number,
-    newOrderSide: types.Side,
-    newOrderType: types.OrderType,
-    newOrderTag: String = constants.DEFAULT_ORDER_TAG
+    ClientOrderId: number,
+    Price: number,
+    Size: number,
+    Side: types.Side,
+    Tag: String = constants.DEFAULT_ORDER_TAG
   ): Promise<TransactionSignature> {
     return await this.getSubClient(asset).immediateOrCancel(
       this.marketIdentifierToPublicKey(asset, market),
-      newOrderClientOrderId,
-      newOrderPrice,
-      newOrderSize,
-      newOrderSide,
-      newOrderType,
-      newOrderTag
+      ClientOrderId,
+      Price,
+      Size,
+      Side,
+      Tag
     );
   }
 
--- a/src/client.ts
+++ b/src/client.ts
@@ -555,19 +555,19 @@ export class Client {
   public async immediateOrCancel(
     asset: Asset,
     market: PublicKey,
-    ClientOrderId: number,
-    Price: number,
-    Size: number,
-    Side: types.Side,
-    Tag: String = constants.DEFAULT_ORDER_TAG
+    clientOrderId: number,
+    price: number,
+    size: number,
+    side: types.Side,
+    tag: String = constants.DEFAULT_ORDER_TAG
   ): Promise<TransactionSignature> {
     return await this.getSubClient(asset).immediateOrCancel(
       this.marketIdentifierToPublicKey(asset, market),
-      ClientOrderId,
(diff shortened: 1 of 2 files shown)


src/subclient.ts
```diff
@@ -1176,6 +1176,56 @@ export class SubClient {
...
+      let marketIndex = this._subExchange.markets.getMarketIndex(market);
+      tx.add(
+        instructions.placeOrderV3Ix(
+          this.asset,
+          marketIndex,
+          newOrderPrice,
+          newOrderSize,
+          newOrderSide,
+          newOrderType,
+          newOrderClientOrderId,
```
filipzeta, line 1208: newOrderClientOrderId will have to be something reserved, because if someone else has used the same one then the cancel might not work right? maybe we just hardcode u64 max or something instead of letting the user pass it in

CC @xjcaa i remember we chatted about this at some point

src/subclient.ts
```diff
@@ -1176,6 +1176,56 @@ export class SubClient {
...
+   * @param newOrderClientOrderId   the subClient order id for the new order
+   * @param newOrderPrice           the native price of the order (6 d.p) as integer
+   * @param newOrderSize            the quantity of the order (3 d.p) as integer
+   * @param newOrderSide            the side of the order. bid / ask
+   * @param newOrderType            the type of the order, limit / ioc / post-only
+   * @param newOrderTag     optional: the string tag corresponding to who is inserting. Default "SDK", max 4 length
+   */
+     public async immediateOrCancel(
+      market: PublicKey,
+      newOrderClientOrderId: number,
```
filipzeta, line 1191: nit: i think these should just be clientOrderId, price, etc. like in placeOrder() or cancelOrder() because there isn't a 'new' order here, it's just a single order. we use 'new' in the replace functions when there's an old and new order

src/subclient.ts
```diff
@@ -1176,6 +1176,56 @@ export class SubClient {
     return await utils.processTransaction(this._parent.provider, tx);
   }
 
+    /**
+   * Sends an IOC by sending a placeOrder follow by a no error cancel to cancel the remaining balance
+   * @param market                  the market address of the order to be cancelled and new order.
+   * @param newOrderClientOrderId   the subClient order id for the new order
```
filipzeta, line 1182: lol i just realised this typo - `the subClient order id` should be `the client order id`, most likely from my find-and-replace when doing the multiasset PR 😝 ill make a typo-fixing PR after we merge this

src/subclient.ts
epsilonsized: ah that makes sense thank you

src/subclient.ts
epsilonsized: cool, I'll leave this in then!

src/subclient.ts
epsilonsized: talked to @xjcaa makes sense to hardcode it - would I be able to get a suggested value?

filipzeta: I'd use the maximum for `number` which is 2^53 - 1, or something really large and specific but under that limit like 999999999999999. Put it in constants.ts btw

@xjcaa do we want to add a quick check to functions like placeOrder to verify that the clientOrderId provided is not the same as our reserved IoC one?

also sidenote what happens if another user inserts an order with the same clientOrderId that we do an IoC for? will the IoC just fail to cancel?

src/client.ts
```diff
@@ -555,21 +555,19 @@ export class Client {
   public async immediateOrCancel(
     asset: Asset,
     market: PublicKey,
-    newOrderClientOrderId: number,
-    newOrderPrice: number,
-    newOrderSize: number,
-    newOrderSide: types.Side,
-    newOrderType: types.OrderType,
-    newOrderTag: String = constants.DEFAULT_ORDER_TAG
+    ClientOrderId: number,
```
filipzeta, line 558: sorry ive got another nitpicky one but these variable names should be lowercase, like `clientOrderId`, `price`, etc

epsilonsized: thanks

PR #149, opened 2022-08-30
Bugfix client subscriptions

filipzeta:
If we use `callback` in `subclient.load()`, only the last loaded subclient's callback will actually fire on trade/ordercomplete events because of some Anchor shenanigans.

This PR moves that subscription logic out to the client to fix this issue for 1 client. I'll raise an Anchor PR too.

### DIFF
--- a/src/events.ts
+++ b/src/events.ts
@@ -57,6 +57,8 @@ export function eventTypeToString(event: EventType) {
       return "GREEKS";
     case EventType.TRADE:
       return "TRADE";
+    case EventType.ORDERCOMPLETE:
+      return "ORDERCOMPLETE";
     case EventType.ORDERBOOK:
       return "ORDERBOOK";
     case EventType.ORACLE:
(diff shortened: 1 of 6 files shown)


src/events.ts
```diff
@@ -57,6 +57,8 @@ export function eventTypeToString(event: EventType) {
       return "GREEKS";
     case EventType.TRADE:
       return "TRADE";
+    case EventType.ORDERCOMPLETE:
```
filipzeta, line 60: We just forgot this in a previous PR 😝

xjcaa (approved):
LGTM

src/program-types.ts
```diff
@@ -268,8 +268,8 @@ export interface PlaceOrderEvent {
 export interface TradeEvent {
   marginAccount: PublicKey;
   index: number;
-  costOfTrades: anchor.BN;
   size: anchor.BN;
+  costOfTrades: anchor.BN;
```
filipzeta, line 272: Aligning with events.rs

PR #152, opened 2022-09-14
Perps

filipzeta:


examples/cranking/cranking-example.ts
```diff
@@ -29,7 +29,7 @@ switch (process.env["network"]) {
 
 const NETWORK_URL = process.env["network_url"]!;
 const PROGRAM_ID = new PublicKey(process.env["program_id"]);
-let crankingMarkets = new Array(constants.ACTIVE_MARKETS).fill(false);
+let crankingMarkets = new Array(constants.ACTIVE_MARKETS - 1).fill(false);
```
xjcaa, line 32: is this right? active markets is 46 right, total markets is 138?

filipzeta: I added + 1 to ACTIVE_MARKETS for perps

src/utils.ts
```diff
@@ -708,6 +723,7 @@ export async function processTransaction(
     );
     return txSig;
   } catch (err) {
+    console.log(err);
```
filipzeta, line 726: - [ ] remove before merging

examples/cranking/cranking-example.ts
```diff
@@ -29,7 +29,7 @@ switch (process.env["network"]) {
 
 const NETWORK_URL = process.env["network_url"]!;
```
xjcaa, line 30: This also needs to crank the perp market?

src/client.ts
```diff
@@ -357,14 +358,14 @@ export class Client {
...
-        Exchange.getSubExchange(asset).markets.markets[market].address;
+      marketPubkey = Exchange.getMarket(asset, market).address;
     } else {
       marketPubkey = market;
     }
     return marketPubkey;
   }
 
+  // TODO giving Exchange.getPerpMarket(asset).marketIndex doesn't work here, fix
   public async placeOrder(
```
xjcaa, line 369: I think just use a separate api man - > placePerpOrder(asset, ...)

src/constants.ts
```diff
@@ -26,7 +27,7 @@ export const MAX_SET_REFERRALS_REWARDS_ACCOUNTS = 12;
 export const MARKET_INDEX_LIMIT = 18;
 // 3 accounts per set * 9 = 27 + 2 = 29 accounts.
 export const CLEAN_MARKET_LIMIT = 9;
-export const CRANK_ACCOUNT_LIMIT = 12;
+export const CRANK_ACCOUNT_LIMIT = 10;
```
xjcaa, line 30: can we have diff limit for perp and normal

src/risk.ts
```diff
@@ -19,18 +19,27 @@ import {
```
xjcaa, line 19: This file should handle the net funding pnl of the position vs greeks as wel

kelvinlau20100 (approved):
lfg

PR #157, opened 2022-09-16
Add cancelAllMarketOrders

filipzeta:
This PR adds a new instruction `cancelAllMarketOrders()` that will cancel all orders on a single market index in one go using Serum's `cpi::prune()`. The only other difference between `cancelAllMarketOrders()` and `cancelOrder()` is that here we don't provide a side + order ID. 

zeta-options PR: https://github.com/zetamarkets/zeta-options/pull/472
Fuze PR: https://github.com/zetamarkets/fuze/pull/28

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,12 +5,9 @@ Version changes are pinned to SDK releases.
 
 ## [Unreleased]
 
-## [0.16.19]
-
-- client: add cancelAllMarketOrders(). ([#157](https://github.com/zetamarkets/sdk/pull/157))
-
 ## [0.16.18]
 
+- client: add cancelAllMarketOrders(). ([#157](https://github.com/zetamarkets/sdk/pull/157))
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -5,12 +5,18 @@ Version changes are pinned to SDK releases.
 
 ## [Unreleased]
 
+## [0.16.19]
+
+- client: add cancelAllMarketOrders(). ([#157](https://github.com/zetamarkets/sdk/pull/157))
```
kelvinlau20100, line 10: did 0.16.18 get released?

kelvinlau20100 (approved):
lgtm except version looks strange, i dont think we've released any of it right?

CHANGELOG.md
filipzeta: ooo good point, checked npm and 0.16.18 isnt released so ill lump this into that version

PR #163, opened 2022-10-07
Fix Kind enum

filipzeta:
Broke devnet due to IDL decoding, this fixes it. Mainnet is fine either way, having unused values in the enum doesn't cause any problems.

Already released as 0.16.19 yesterday.


kelvinlau20100 (approved):
lgtm

PR #166, opened 2022-10-14
Add optional blockhash option to placeOrder() + bump examples

filipzeta:
Calculating the blockhash involves a roundtrip to the RPC, which takes time. SDK users can lower their latency by pre-caching the blockhash instead. This PR adds an optional `blockhash?: string` to `utils.processTransaction()`. It's only used in `client.placeOrder()` for now but we can theoretically add it to any transaction if there's appetite for it.

This PR also bumps the examples - I noticed that the SDK version and webserver URL were out of date 😊

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@ Version changes are pinned to SDK releases.
 
 ## [Unreleased]
 
+## [0.17.1]
+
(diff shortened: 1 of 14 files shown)


kelvinlau20100 (approved):
lgtm

xjcaa (approved):

PR #171, opened 2022-11-02
Change asset enum value.

xjcaa:


src/assets.ts
```diff
@@ -79,3 +79,18 @@ export function fromProgramAsset(asset: any): Asset {
   }
   throw Error("Invalid asset");
 }
+
+export function indexToAsset(index: number): Asset {
```
filipzeta, line 83: Is this used anywhere?

filipzeta (approved):
Looks good. Will just need to be careful when bumping server + webserver + indexers, let's do it after expiry.

PR #176, opened 2022-12-08
TIF orders

kelvinlau20100:


src/exchange.ts
```diff
@@ -420,12 +420,6 @@ export class Exchange {
       })
     );
 
-    await Promise.all(
```
xjcaa, line 423: i have context but for @filipzeta - good to comment why

src/exchange.ts
```diff
@@ -570,9 +564,20 @@ export class Exchange {
...
   public async updateAllOrderbooks(live: boolean = true) {
     // This assumes that every market has 1 asksAddress and 1 bidsAddress
-    let allLiveMarkets = this._markets;
+    let allLiveMarkets = [];
+    this.assets.map(async (asset) => {
+      allLiveMarkets = allLiveMarkets.concat(this.getMarkets(asset));
+    });
+
+    for (let i = 0; i < allLiveMarkets.length; i++) {
+      console.log(
```
xjcaa, line 573: todo: remove this later

src/exchange.ts
```diff
@@ -570,9 +564,20 @@ export class Exchange {
 
   public async updateAllOrderbooks(live: boolean = true) {
     // This assumes that every market has 1 asksAddress and 1 bidsAddress
-    let allLiveMarkets = this._markets;
+    let allLiveMarkets = [];
+    this.assets.map(async (asset) => {
+      allLiveMarkets = allLiveMarkets.concat(this.getMarkets(asset));
```
xjcaa, line 569: doesn't this create a copy or is that a ref @kelvinlau20100

src/exchange.ts
filipzeta: is it because subExchange already holds markets?

kelvinlau20100: so concat actually creates a shallow copy of the markets, now that we actually need up to date market state in some of our utils and when creating the orderbook off chain we need to actually use the markets held in subExchange and not the copy

src/exchange.ts
kelvinlau20100: fine in this situation but have updated code slightly

src/client.ts
```diff
@@ -367,31 +367,23 @@ export class Client {
     price: number,
     size: number,
     side: types.Side,
-    type: types.OrderType = types.OrderType.LIMIT,
-    clientOrderId = 0,
-    tag: String = constants.DEFAULT_ORDER_TAG,
-    blockhash?: string
+    options: types.OrderOptions = types.defaultOrderOptions()
```
filipzeta, line 370: just noting that this seems to be a breaking change and so should be highlighted in the SDK changelogs

src/utils.ts
```diff
@@ -1630,3 +1645,50 @@ export function getProductLedger(marginAccount: MarginAccount, index: number) {
   }
   return marginAccount.productLedgers[index];
 }
+
+/*
+ TODO: Maybe this should all be done in BN arithmetic because
```
filipzeta, line 1650: how large?

src/client.ts
```diff
@@ -401,19 +393,13 @@ export class Client {
...
   ): Promise<TransactionSignature> {
-    return await this.getSubClient(asset).placePerpOrder(
+    return await this.getSubClient(asset).placePerpOrderV2(
```
xjcaa, line 398: should we also make it placePerp and placeOrder

src/exchange.ts
xjcaa: wait why is it fine?

xjcaa: cos from what you told me, the markets in allLiveMarkets are copies, so the update orderbook updates the copy, not the original? but if it works then some part of the story is incorrect right

src/subclient.ts
```diff
@@ -610,15 +613,12 @@ export class SubClient {
    * from the user open orders account before cancelling the second order.
    * (Depending on the order in which the order was cancelled).
    */
-  public async placeOrderV3(
+  public async placeOrderV4(
```
xjcaa, line 616: should we nuke vX and just make it placeOrder?

src/subclient.ts
```diff
@@ -687,14 +698,11 @@ export class SubClient {
    * from the user open orders account before cancelling the second order.
    * (Depending on the order in which the order was cancelled).
    */
-  public async placePerpOrder(
+  public async placePerpOrderV2(
```
xjcaa, line 701: just call it placePerpOrder, but make ix v2?

src/utils.ts
xjcaa: yea i feel like it should?

kelvinlau20100: i gave it more thought i think the only one you would need to do this for is the seqNum comparison, timestamps seem fine to leave as number? but maybe better for consistency

filipzeta (approved):

PR #178, opened 2022-12-20
Post only slide

kelvinlau20100:


src/serum/market.ts
```diff
@@ -492,7 +492,10 @@ function divideBnToNumber(numerator: BN, denominator: BN): number {
   const quotient = numerator.div(denominator).toNumber();
   const rem = numerator.umod(denominator);
   const gcd = rem.gcd(denominator);
-  return quotient + rem.div(gcd).toNumber() / denominator.div(gcd).toNumber();
+  return +(
```
xjcaa, line 495: what this for

xjcaa (approved):

src/serum/market.ts
kelvinlau20100: oh floating arithmetic lead to non exact integers with like 10d.p which was messing with tests

PR #181, opened 2022-12-22
filter client orders

kelvinlau20100:


src/subclient.ts
```diff
@@ -1539,7 +1539,25 @@ export class SubClient {
       })
     );
 
-    this._orders = [].concat(...orders);
+    let allOrders = [].concat(...orders);
+    const asset = this._asset;
+    this._orders = allOrders.filter(function (order: types.Order) {
+      let seqNum = utils.getSeqNumFromSerumOrderKey(
+        order.orderId,
+        order.side == types.Side.BID ? true : false
```
xjcaa, line 1547: Can't this just be side == bid without the ternary

xjcaa (changes requested):

src/subclient.ts
kelvinlau20100: Whoops yes

xjcaa (approved):

PR #182, opened 2022-12-23
Refactor tif offset

xjcaa:


src/types.ts
```diff
@@ -299,18 +299,25 @@ export function fromProgramOrderCompleteType(
...
-  explicitTIF?: boolean;
-  tifOffset?: number;
+  tifOptions: TIFOptions;
   orderType?: types.OrderType;
   clientOrderId?: number;
   tag?: string;
   blockhash?: string;
 }
 
+/* Only set one of these values.
```
kelvinlau20100, line 309: old comment

src/types.ts
```diff
@@ -299,18 +299,25 @@ export function fromProgramOrderCompleteType(
...
+ * expiryOffset - seconds in future that the order will expire.
+ */
+export interface TIFOptions {
+  expiryOffset?: number | undefined;
```
kelvinlau20100, line 313: this should be defined if its the only one

src/utils.ts
```diff
@@ -1646,27 +1646,28 @@ export function getProductLedger(marginAccount: MarginAccount, index: number) {
...
-  tifOffset: number
-) {
-  if (explicitTIF) {
-    return tifOffset;
+export function getTIFOffset(marketInfo: Market, tifOptions: types.TIFOptions) {
+  if (tifOptions.expiryOffset == 0) {
+    throw new Error("Invalid expiry offset");
+  }
+
+  if (tifOptions.expiryOffset == undefined) {
```
kelvinlau20100, line 1654: yeah im not that set on making this defined its up to you

src/utils.ts
```diff
@@ -1646,27 +1646,28 @@ export function getProductLedger(marginAccount: MarginAccount, index: number) {
...
+    return 0;
   }
 
   let currEpochStartTs = marketInfo.serumMarket.epochStartTs.toNumber();
   let epochLength = marketInfo.serumMarket.epochLength.toNumber();
-
+  let epochEnd = currEpochStartTs + epochLength;
   let now = Exchange.clockTimestamp;
+  let desiredExpiryTs = now + tifOptions.expiryOffset;
+  let desiredOffset = desiredExpiryTs % epochLength;
```
kelvinlau20100, line 1663: i think we can return at this modulo?

xjcaa: doesnt accomodate wrap

src/types.ts
xjcaa: discussed offline

PR #185, opened 2023-01-05
LUT

kelvinlau20100:


src/subclient.ts
```diff
@@ -1542,12 +1573,14 @@ export class SubClient {
     this.delegatedCheck();
     let tx = this.getPositionMovementTx(movementType, movements);
     let response = await utils.simulateTransaction(this._parent.provider, tx);
-
-    let events = response.events;
     let positionMovementEvent = undefined;
-    for (var i = 0; i < events.length; i++) {
-      if (events[i].name == "PositionMovementEvent") {
-        positionMovementEvent = events[i].data;
+    while (true) {
```
filipzeta, line 1577: whats the purpose of this change?

src/utils.ts
```diff
@@ -691,43 +690,63 @@ export async function processTransaction(
...
+    }
 
-  if (useLedger) {
-    tx = await Exchange.ledgerWallet.signTransaction(tx);
+    let v0Tx: VersionedTransaction = new VersionedTransaction(
+      new TransactionMessage({
+        payerKey: provider.wallet.publicKey,
+        recentBlockhash:
+          blockhash ??
+          (await provider.connection.getRecentBlockhash()).blockhash,
```
filipzeta, line 708: getRecentBlockhash should change to `getLatestBlockhash(commitmentConfig("finalized"))` as per https://github.com/zetamarkets/sdk/pull/193

src/utils.ts
```diff
@@ -691,43 +690,63 @@ export async function processTransaction(
...
+        })
+    );
+    v0Tx = (await provider.wallet.signTransaction(
+      v0Tx
+    )) as VersionedTransaction;
+    rawTx = v0Tx.serialize();
   } else {
-    tx = await provider.wallet.signTransaction(tx);
+    tx.recentBlockhash =
+      blockhash ?? (await provider.connection.getRecentBlockhash()).blockhash;
```
filipzeta, line 725: finalized blockhash here too pls :)

src/subclient.ts
kelvinlau20100: new version of anchor has changed the event parsing

src/utils.ts
kelvinlau20100: yep apologies

xjcaa:
what's the impact of https://github.com/coral-xyz/anchor/pull/2018/files?

src/program-types.ts
```diff
@@ -145,7 +145,7 @@ export interface Position {
 
 export interface OrderState {
   closingOrders: anchor.BN;
-  openingOrders: [anchor.BN, anchor.BN];
+  openingOrders: Array<anchor.BN>;
```
xjcaa, line 148: was there an error for this

src/utils.ts
```diff
@@ -726,19 +705,67 @@ export async function processTransaction(
     );
   }
 
-  if (useLedger) {
-    tx = await Exchange.ledgerWallet.signTransaction(tx);
+  if (useVersioned) {
```
xjcaa, line 708: is there a reason to useVersioned apart from LUT? - if not, could we assume if LUT is not undefined then we use versioned?

xjcaa (approved):

src/program-types.ts
kelvinlau20100: yes

src/utils.ts
kelvinlau20100: yes i think that makes sense

src/utils.ts
```diff
@@ -692,31 +692,10 @@ export async function processTransaction(
   signers?: Array<Signer>,
   opts?: ConfirmOptions,
   useLedger: boolean = false,
+  lutAccs?: AddressLookupTableAccount[],
```
xjcaa, line 695: default to undefined?

kelvinlau20100: can't default value it because it's an optional argument

PR #187, opened 2023-01-10
Delegated accounts

filipzeta:
Allow the user to specify a separate pubkey on behalf of which it performs actions - currently only placing and cancelling orders. We can add spread accounts and liquidations later because this is enough for now IMO.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 All notable changes to this project will be documented in this file.
 Version changes are pinned to SDK releases.
 
+## [0.19.4] 2023-01-18
+
+- general: Add delegated account functionality to program and SDK. ([#187](https://github.com/zetamarkets/sdk/pull/187))
+
 ## [0.19.3] 2023-01-09
 
 - client: add perp functionality to all cancel+place functions in subclient. ([#186](https://github.com/zetamarkets/sdk/pull/186))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 All notable changes to this project will be documented in this file.
 Version changes are pinned to SDK releases.
 
-## [0.19.4] 2023-01-18
+## [0.20.0] 2023-01-18
 
 - general: Add delegated account functionality to program and SDK. ([#187](https://github.com/zetamarkets/sdk/pull/187))
 
--- a/src/client.ts
+++ b/src/client.ts
@@ -367,7 +367,8 @@ export class Client {
     price: number,
     size: number,
     side: types.Side,
-    options: types.OrderOptions = types.defaultOrderOptions()
+    options: types.OrderOptions = types.defaultOrderOptions(),
+    onBehalfOfMarginAccountAddress: PublicKey = undefined
   ): Promise<TransactionSignature> {
     let marketPubkey = this.marketIdentifierToPublicKey(asset, market);
     if (marketPubkey == Exchange.getPerpMarket(asset).address) {
@@ -375,7 +376,8 @@ export class Client {
         price,
         size,
         side,
-        options
+        options,
+        onBehalfOfMarginAccountAddress
       );
     } else {
       return await this.getSubClient(asset).placeOrder(
@@ -393,13 +395,15 @@ export class Client {
     price: number,
     size: number,
     side: types.Side,
-    options: types.OrderOptions = types.defaultOrderOptions()
+    options: types.OrderOptions = types.defaultOrderOptions(),
+    onBehalfOfMarginAccountAddress: PublicKey = undefined
   ): Promise<TransactionSignature> {
     return await this.getSubClient(asset).placePerpOrder(
       price,
       size,
       side,
-      options
+      options,
+      onBehalfOfMarginAccountAddress
     );
   }
 
@@ -455,6 +459,13 @@ export class Client {
     return this.getSubClient(asset).createCancelAllMarketOrdersInstruction(
       this.marketIdentifierToIndex(asset, market)
     );
+  }
+
+  public async editDelegatedPubkey(
+    asset: Asset,
+    delegatedPubkey: PublicKey
+  ): Promise<TransactionSignature> {
+    return await this.getSubClient(asset).editDelegatedPubkey(delegatedPubkey);
   }
 
   public async migrateFunds(
--- a/src/client.ts
+++ b/src/client.ts
@@ -144,7 +144,8 @@ export class Client {
     wallet: types.Wallet,
     opts: ConfirmOptions = utils.defaultCommitment(),
     callback: (asset: Asset, type: EventType, data: any) => void = undefined,
-    throttle: boolean = false
+    throttle: boolean = false,
+    onBehalfOfUser: PublicKey = undefined
   ): Promise<Client> {
     let client = new Client(connection, wallet, opts);
 
@@ -235,11 +236,20 @@ export class Client {
       );
     }
 
+    if (onBehalfOfUser !== undefined) {
+      await client.loadDelegatedClient(onBehalfOfUser);
+    }
+
     await Promise.all(
       client.getAllSubClients().map(async (subclient) => {
         await subclient.updateState();
       })
     );
+    if (onBehalfOfUser !== undefined) {
+      client.delegatedAccount.getAllSubClients().map(async (subclient) => {
+        await subclient.updateState();
+      });
+    }
 
     return client;
   }
@@ -387,8 +397,7 @@ export class Client {
     price: number,
     size: number,
     side: types.Side,
-    options: types.OrderOptions = types.defaultOrderOptions(),
-    onBehalfOfUser: PublicKey = undefined
+    options: types.OrderOptions = types.defaultOrderOptions()
   ): Promise<TransactionSignature> {
     let marketPubkey = this.marketIdentifierToPublicKey(asset, market);
     if (marketPubkey == Exchange.getPerpMarket(asset).address) {
@@ -396,8 +405,7 @@ export class Client {
         price,
         size,
         side,
-        options,
-        onBehalfOfUser
+        options
       );
     } else {
       return await this.getSubClient(asset).placeOrder(
@@ -405,8 +413,7 @@ export class Client {
         price,
         size,
         side,
-        options,
-        onBehalfOfUser
+        options
       );
     }
   }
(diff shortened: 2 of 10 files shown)


src/subclient.ts
```diff
@@ -699,30 +699,89 @@ export class SubClient {
...
       );
-      openOrdersPda = _openOrdersPda;
-      tx.add(initIx);
+      if (
+        (await this._parent.connection.getAccountInfo(openOrdersPda)) == null
+      ) {
+        console.log(
+          `[${assetToName(
+            asset
+          )}] Delegated user doesn't have open orders account. Please make one and retry. Returning...`
```
filipzeta, line 744: Delegated user can't create the open orders account because the actual user has to sign it

src/subclient.ts
```diff
@@ -699,30 +699,89 @@ export class SubClient {
...
-      );
+    let tx = new Transaction();
+    let market = Exchange.getPerpMarket(asset).address;
 
-      let [initIx, _openOrdersPda] = await instructions.initializeOpenOrdersIx(
-        this.asset,
-        market,
-        this._parent.publicKey,
-        this.marginAccountAddress
+    if (onBehalfOfMarginAccountAddress != undefined) {
```
filipzeta, line 717: Is this whole loading section messy? Can't just load a new client because we don't have the private key of the `onBehalfOfMarginAccountAddress`, so we have to load a few bits and pieces manually. Could maybe split off into a different function but not sure if it'll be much cleaner?

src/client.ts
```diff
@@ -626,18 +642,33 @@ export class Client {
...
     let tx = new Transaction();
     let index = Exchange.getZetaGroupMarkets(asset).getMarketIndex(
       this.marketIdentifierToPublicKey(asset, market)
     );
+    let subClient = this.getSubClient(asset);
+    let marginAccountAddress = subClient.marginAccountAddress;
+    let openOrders = subClient.openOrdersAccounts[index];
+
+    if (onBehalfOfUser != undefined) {
+      let delegatedClient = await Client.load(
```
xjcaa, line 657: this load in itself would take way too long for this to be UX friendly each time

xjcaa: you should be able to load a delegated client directly imo

src/subclient.ts
```diff
@@ -618,29 +618,56 @@ export class SubClient {
...
-        `[${assetToName(
-          this.asset
-        )}] User doesn't have open orders account. Initialising for market ${market.toString()}.`
+    let asset = this._asset;
+    let marginAccountAddress = this.marginAccountAddress;
+    let whitelistTradingFeesAddress = this._parent.whitelistTradingFeesAddress;
+    let openOrders = null;
+
+    if (onBehalfOfUser != undefined) {
+      let delegatedClient = await Client.load(
```
xjcaa, line 632: yeah, i don't think this inter tx client load makes sense?

src/subclient.ts
xjcaa: should we let the delegated user create an open orders account?

src/client.ts
```diff
@@ -230,11 +236,20 @@ export class Client {
...
+    if (onBehalfOfUser !== undefined) {
+      await client.loadDelegatedClient(onBehalfOfUser);
+    }
+
     await Promise.all(
       client.getAllSubClients().map(async (subclient) => {
         await subclient.updateState();
       })
     );
+    if (onBehalfOfUser !== undefined) {
```
xjcaa, line 248: it doesn't make sense to load any state of the delegatee if they decide to pass in the `onBehalfOfUser` in the SDK

src/client.ts
```diff
@@ -139,7 +144,8 @@ export class Client {
     wallet: types.Wallet,
     opts: ConfirmOptions = utils.defaultCommitment(),
     callback: (asset: Asset, type: EventType, data: any) => void = undefined,
-    throttle: boolean = false
+    throttle: boolean = false,
+    onBehalfOfUser: PublicKey = undefined
```
xjcaa, line 148: this should be called `delegator`.

src/types.ts
```diff
@@ -32,6 +32,26 @@ export class DummyWallet implements Wallet {
   }
 }
 
+export class DelegatedWallet implements Wallet {
```
xjcaa, line 35: why does this need to exist?

src/subclient.ts
```diff
@@ -660,10 +682,10 @@ export class SubClient {
       options.clientOrderId != undefined ? options.clientOrderId : 0,
       options.tag != undefined ? options.tag : constants.DEFAULT_ORDER_TAG,
       tifOffsetToUse,
-      this.marginAccountAddress,
+      marginAccountAddress,
       this._parent.publicKey,
-      openOrdersPda,
-      this._parent.whitelistTradingFeesAddress
+      openOrders,
+      whitelistTradingFeesAddress
```
xjcaa, line 688: i.e. feels like this should just have the logic to pass in a separate authority if delegated client

src/client.ts
```diff
@@ -252,6 +267,21 @@ export class Client {
     // This is referring itself by another referrer.
   }
 
+  public async loadDelegatedClient(onBehalfOfUser: PublicKey) {
```
xjcaa, line 270: feels like we don't need this, just a client that is returned if delegator is passed in

xjcaa:
add errors on using non allowed delegated functions?

src/types.ts
filipzeta: need to create a wallet that just has a pubkey, no secret key. is there an easier way?

src/client.ts
```diff
@@ -139,10 +147,16 @@ export class Client {
     wallet: types.Wallet,
     opts: ConfirmOptions = utils.defaultCommitment(),
     callback: (asset: Asset, type: EventType, data: any) => void = undefined,
-    throttle: boolean = false
+    throttle: boolean = false,
+    delegator: PublicKey = undefined
   ): Promise<Client> {
     let client = new Client(connection, wallet, opts);
 
+    if (delegator != undefined) {
```
xjcaa, line 155: like here for delegated wallet, can't you just do

clientPubkey = delegator, and sub that in for wallet.publicKey? - I see you pass it into subclient too

src/client.ts
```diff
@@ -139,10 +147,16 @@ export class Client {
     wallet: types.Wallet,
     opts: ConfirmOptions = utils.defaultCommitment(),
     callback: (asset: Asset, type: EventType, data: any) => void = undefined,
-    throttle: boolean = false
+    throttle: boolean = false,
+    delegator: PublicKey = undefined
   ): Promise<Client> {
     let client = new Client(connection, wallet, opts);
```
xjcaa, line 153: like assuming this wallet is the delegatee's wallet -> the client.provider.wallet.publicKey is the wrong one or in subclient -> this._parent.provider -> referring to wrong the wallet?

src/subclient.ts
```diff
@@ -391,7 +392,7 @@ export class SubClient {
         instructions.initializeMarginAccountIx(
           this._subExchange.zetaGroupAddress,
           this._marginAccountAddress,
-          this._parent.publicKey
+          this._parent.provider.wallet.publicKey
```
xjcaa, line 395: maybe functionally it makes to overhaul a bit of the logic such that we use universally ->

`client.ownerKey` -> as the pubkey that owns the margin accounts
`client.signerKey` or something similar as the key that is signing on behalf the tx, that way we're not like using

wallet.publicKey
provider.wallet.publicKey etc across everything

src/client.ts
```diff
@@ -474,7 +498,7 @@ export class Client {
         amount,
         withdrawSubClient.marginAccountAddress,
         this.usdcAccountAddress,
-        this.publicKey
```
xjcaa, line 477: fwiw. this.publicKey is the delegatee's publicKey which is still the same as this.provider.wallet.publicKey so this change functionally doesn't do anything right?

src/subclient.ts
xjcaa: Yeah reading over it should be consistent, like if we are trying to pass in the key for signing (owner or delegator) -> we use something explictly different to owner

xjcaa:
lookin alot nicer tho nearly there

src/client.ts
```diff
@@ -1059,4 +1095,12 @@ export class Client {
       this._orderCompleteEventListener = undefined;
     }
   }
+
+  private delegatedCheck() {
+    if (this.delegatorKey) {
+      throw Error(
```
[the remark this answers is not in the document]
filipzeta: we do `throw Error` in loads of places already
image

also just tried it locally and i see the error fine

src/client.ts
filipzeta: yep good call, im removing DelegatedWallet because we actually dont need it

src/client.ts
filipzeta: yea ive changed it to be consistently provider.publicKey for the auth/signer because thatll change for delegated accounts

src/client.ts
filipzeta: i made it such that the provider is always the one that signs, so client.provider.wallet.publicKey will be the delegated account whereas client.publicKey will be the delegatee.

if its just a standard client load then they'll both be the same

CHANGELOG.md
```diff
@@ -3,6 +3,10 @@
 All notable changes to this project will be documented in this file.
 Version changes are pinned to SDK releases.
 
+## [0.19.4] 2023-01-18
+
+- general: Add delegated account functionality to program and SDK. ([#187](https://github.com/zetamarkets/sdk/pull/187))
```
xjcaa, line 8: i reckon move it to unreleased and then bang 0.20.0 with both changes

xjcaa: shud release devnet on beta imo

src/client.ts
```diff
@@ -139,22 +147,26 @@ export class Client {
...
+    let user = wallet.publicKey;
+    if (delegator != undefined) {
+      user = delegator;
+      client._delegatorKey = delegator;
+    }
+
     client._usdcAccountAddress = await utils.getAssociatedTokenAddress(
       Exchange.usdcMintAddress,
-      wallet.publicKey
+      user
```
xjcaa, line 163: nit: i wonder if there is a conflict in the definition of the word user i.e. is it the delegator or the person trading on behalf

xjcaa (approved):
lgtm: can you have one last check that when we generate pdas, it's consistent ref the actual owner of the acc

CHANGELOG.md
filipzeta: yep devnet is 0.20.0-beta.0

src/client.ts
filipzeta: yep agreed, ive changed it to 'owner' because in the context of client.load() its always the actual owner of the margin account

filipzeta:
> lgtm: can you have one last check that when we generate pdas, it's consistent ref the actual owner of the acc

yep, will verify today in devnet alongside oracle stuff and have one last check of pda generation

PR #194, opened 2023-01-24
Add priority fees

xjcaa:


src/exchange.ts
```diff
@@ -230,6 +230,34 @@ export class Exchange {
...
+    microLamportsPerCU: number = constants.DEFAULT_MICRO_LAMPORTS_PER_CU_FEE
+  ) {
+    if (this._usePriorityFees) {
+      throw Error("Priority fees already turned on");
+    }
+    this._usePriorityFees = true;
+    this._priorityFee = microLamportsPerCU;
+  }
+
+  public async updatePriorityFee(microLamportsPerCU: number) {
```
kelvinlau20100, line 254: doesn't need to be async, not optional as well so why do we need to check for undefined?

src/utils.ts
```diff
@@ -715,6 +716,14 @@ export async function processTransaction(
       tx.partialSign(kp);
     });
 
+  if (Exchange.usePriorityFees) {
+    tx.add(
```
kelvinlau20100, line 720: doesn't it need to be at the start of the tx?

filipzeta (approved):

PR #197, opened 2023-02-02
Optimise Exchange loading

kelvinlau20100:
Reduce exchange load times

### DIFF
--- a/src/exchange.ts
+++ b/src/exchange.ts
@@ -265,34 +265,31 @@ export class Exchange {
     this._blockhashCommitment = commitment;
   }
 
-  public async initialize(
-    assets: Asset[],
-    programId: PublicKey,
-    network: Network,
-    connection: Connection,
-    opts: ConfirmOptions,
+  public initialize(
+    loadConfig: types.LoadExchangeConfig,
     wallet = new types.DummyWallet()
   ) {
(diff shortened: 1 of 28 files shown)


src/exchange.ts
```diff
@@ -387,82 +386,119 @@ export class Exchange {
...
     wallet = new types.DummyWallet(),
-    throttleMs = 0,
     callback?: (asset: Asset, event: EventType, data: any) => void
   ) {
     if (this.isInitialized) {
       throw "Exchange already loaded";
     }
 
+    if (loadConfig.network == Network.LOCALNET && loadConfig.loadFromStore) {
+      throw Error("Cannot load localnet from store");
```
xjcaa, line 398: why not? isn't pda the same as devnet

src/serum/market.ts
```diff
@@ -131,18 +132,32 @@ export class Market {
...
+    options: MarketOptions = {},
+    programId: PublicKey,
+    layoutOverride?: any
+  ) {
+    decoded.epochStartTs = new BN(
+      Exchange.clockTimestamp -
+        (Exchange.clockTimestamp % decoded.epochLength.toNumber())
     );
+
+    // console.log(decoded);
```
xjcaa, line 159: log

PR #198, opened 2023-02-03
flex underlyings

xjcaa:


src/assets.ts
```diff
@@ -91,6 +97,9 @@ export function indexToAsset(index: number): Asset {
     case 2: {
       return Asset.ETH;
     }
+    case 3: {
+      return Asset.ETH;
```
kelvinlau20100, line 101: return Asset.APT

src/market.ts
```diff
@@ -220,11 +220,51 @@ export class ZetaGroupMarkets {
...
+    );
+
+    try {
+      instance.updateExpirySeries();
+    } catch (e) {
+      console.log("Test updateExpirySeries");
+      console.log(e);
+    }
+
+    if (perpOnly) {
```
filipzeta, line 264: just to confirm - this will make this._markets an empty list? therefore Exchange.getMarkets() will just return the 1 perps market for perp-only zetaGroups? 

can double check on devnet too, most of our apps rely on those functions and should Just Work™️ if that is the case 😊

src/assets.ts
kelvinlau20100: still stands

src/market.ts
xjcaa: oh it should be [perps] right

src/assets.ts
xjcaa: nuked this function we dont use indices really

PR #201, opened 2023-02-08
modify market store with scripts

xjcaa:
- new method for loading decoded
- use market-gen.ts for generation of json files


src/serum/scripts/market-gen-utils.ts
```diff
@@ -0,0 +1,34 @@
...
+  PublicKey,
+}
+
+export function returnDecodedType(key: string): DecodeType {
+  key = key.valueOf();
+  if (key === "accountFlags") {
+    return DecodeType.AccountFlags;
+  }
+
+  let bnList = [
```
filipzeta, line 13: do we need to remember to update this with any program changes?

xjcaa: never changes but if we do sure

PR #211, opened 2023-03-22
Allow toggling zetaGroup perpsOnly on/off

filipzeta:


src/client.ts
```diff
@@ -420,6 +420,13 @@ export class Client {
     return index;
   }
 
+  // Refresh this._subExchange in the subClients, useful for integration testing
+  public relinkExchange() {
+    for (var subclient of this.subClients.values()) {
+      subclient.updateSubExchange();
```
xjcaa, line 426: is this not async

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@zetamarkets/sdk",
   "repository": "https://github.com/zetamarkets/sdk/",
-  "version": "0.24.1",
+  "version": "0.24.3",
```
xjcaa, line 4: out of datE?

src/client.ts
filipzeta: nah it just does a getSubExchange() which isnt async - the vibe of it is to re-link the subexchange in each subclient after doing an Exchange.load() again

xjcaa (approved):

PR #218, opened 2023-04-04
Allow overrides in unrealised PnL calcs

filipzeta:
On FE you can be shown a positive PnL, but then actually lose money when you go to close that position due to 
1. Taker fees
2. Orderbook spread

This PR accounts for those to hopefully show a more accurate PnL number. FE will feed the webserver's orderbook toplevel into calculateUnrealizedPnl instead of using the mark price


filipzeta:
Some local verification. I'm in a long SOL-PERP position of 4 lots.

image
image

filipzeta:
Some local verification. I'm in a long SOL-PERP position of 4 lots.
image
image

wheeeliam (approved):

PR #219, opened 2023-04-11
Add handy function Exchange.zetaGroupPubkeyToAsset()

filipzeta:
Store a map to grab zetagroup asset by pubkey without an RPC fetch or having to manually filter all zetaGroups

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ Version changes are pinned to SDK releases.
 
 ## Unreleased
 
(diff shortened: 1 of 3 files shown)


filipzeta:
Tested locally with the following code: 
image
This gives the right assets SOL/BTC/ETH/APT/ARB. In reality we wouldn't use `getZetaGroupAddress()`, we'd have the pubkey from elsewhere.

0xtristan (approved):
LGTM thad

PR #229, opened 2023-05-08
Fix risk calcs

filipzeta:
Bit of cleanup in the risk calculator following a bug report

### DIFF
--- a/src/risk.ts
+++ b/src/risk.ts
@@ -259,8 +259,6 @@ export class RiskCalculator {
       return 0;
     }
 
-    console.log(bidOpenOrders, askOpenOrders, size);
-
     let longLots = convertNativeLotSizeToDecimal(bidOpenOrders);
     let shortLots = convertNativeLotSizeToDecimal(askOpenOrders);
 
--- a/src/risk.ts
+++ b/src/risk.ts
@@ -249,74 +249,62 @@ export class RiskCalculator {
   ): number {
     let asset = fromProgramAsset(marginAccount.asset);
     let marketMaker = types.isMarketMaker(marginAccount);
-    let margin = 0;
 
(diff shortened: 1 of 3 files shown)


src/risk.ts
```diff
@@ -249,74 +249,62 @@ export class RiskCalculator {
...
-      let shortLotsMarginReq = this.getMarginRequirement(
-        asset,
-        i == ledgers.length - 1 ? constants.PERP_INDEX : i,
-        // Negative for sells.
-        -shortLots,
-        types.MarginType.INITIAL
-      );
-      if (
-        (i + 1) % constants.PRODUCTS_PER_EXPIRY == 0 ||
-        i == constants.PERP_INDEX
```
filipzeta, line 294: pretty sure this check was causing issues because 0 <= i <= 46 here, as we just concatenate the perp market to the end of the list

src/risk.ts
```diff
@@ -249,74 +249,62 @@ export class RiskCalculator {
...
+    }
 
-      if (!marketMaker || skipConcession) {
-        if (size > 0) {
-          longLots += Math.abs(convertNativeLotSizeToDecimal(size));
-        } else if (size < 0) {
-          shortLots += Math.abs(convertNativeLotSizeToDecimal(size));
-        }
-      }
+    console.log(bidOpenOrders, askOpenOrders, size);
```
wheeeliam, line 262: intentional?

filipzeta: accidentally left it in from debugging, removed 😊

wheeeliam (approved):

src/risk.ts
0xtristan: oh lol

PR #230, opened 2023-05-12
Migrate vaults

filipzeta:


filipzeta:
- [ ] Remove old insuranceVault, vault and socializedLoss from subexchange when we're happy with everything and remove the 'old' ixs

src/insurance-client.ts
```diff
@@ -128,6 +118,17 @@ export class InsuranceClient {
     return insuranceClient;
   }
 
+  public async reloadInsuranceDepositAccount(): Promise<void> {
+    let [insuranceDepositAccountAddress, _insuranceDepositAccountNonce] =
+      await utils.getUserInsuranceDepositAccount(
```
kelvinlau20100, line 123: this doesn't need to be async

kelvinlau20100 (approved):
lgtm one comment but also might be cleanup based on small changes in vault migration PR

PR #235, opened 2023-06-07
Use confirmTransaction with new args

filipzeta:
Using `confirmTransaction()` with the new `BlockheightBasedTransactionConfirmationStrategy` provides a better UX in regards to transactions that time out or take a while to confirm. This PR changes that, and there's some extra syntax bits regarding the optional `blockhash` argument

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ Version changes are pinned to SDK releases.
 
 ## Unreleased
 
+## [0.28.1] 2023-06-07
+
+- utils: use confirmTransaction() with new BlockheightBasedTransactionConfirmationStrategy. ([#235](https://github.com/zetamarkets/sdk/pull/235))
+
 ## [0.28.0] 2023-05-24
 
 - general: Migrate per-asset insurance vault, deposit vault and socialized loss account to be global accounts. ([#230](https://github.com/zetamarkets/sdk/pull/230))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ Version changes are pinned to SDK releases.
 
 ## [0.28.1] 2023-06-07
 
-- utils: use confirmTransaction() with new BlockheightBasedTransactionConfirmationStrategy. ([#235](https://github.com/zetamarkets/sdk/pull/235))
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,10 @@ Version changes are pinned to SDK releases.
 
 ## Unreleased
 
+## [0.28.1] 2023-06-07
+
+- utils: use confirmTransaction() with new BlockheightBasedTransactionConfirmationStrategy. ([#235](https://github.com/zetamarkets/sdk/pull/235))
```
xjcaa, line 10: yeah what does it do though

filipzeta: allows you to send the block height after which the tx failed, instead of just relying on N seconds timeout. i added an extra bit to the changelog to explain that

src/exchange.ts
```diff
@@ -273,6 +273,8 @@ export class Exchange {
 
   private _programSubscriptionIds: number[] = [];
 
+  private _eventEmitters: any[] = [];
```
filipzeta, line 276: accidentally removed in previous pr when merging main

kelvinlau20100 (approved):

src/utils.ts
```diff
@@ -825,11 +819,20 @@ export async function processTransaction(
...
+    let result = await provider.connection.confirmTransaction({
+      blockhash: recentBlockhash.blockhash,
+      lastValidBlockHeight: recentBlockhash.lastValidBlockHeight,
+      signature: txSig,
+    });
+    if (result.value.err != null) {
+      let parsedErr = parseError(result.value.err);
+      throw parsedErr;
+    }
+    return txSig;
```
MeanMangosteen, line 835: Why don't we use 
```
 export function sendAndConfirmRawTransaction(
    connection: Connection,
    rawTransaction: Buffer,
    options?: ConfirmOptions,
  ): Promise<TransactionSignature>;
  ```
  here?

src/exchange.ts
MeanMangosteen: Curious, the type can't be narrowed here?

src/utils.ts
filipzeta: it doesn't have the `strategy` argument, which is all we're really adding in this PR

src/exchange.ts
filipzeta: i think it can we've just been lazy :D ill add it in the cross margin account stuff

PR #236, opened 2023-06-08
Use automatic priority fees

filipzeta:
Allow the user to toggle on automatic priority fees, which polls getRecentPrioritizationFees() to determine a good fee using a median of the last 20 blocks of fees. Note that we pass in the perpsyncqueue and greeks accounts to the RPC function, as they're the most touched accounts during trading.

We also refactor the existing priority fee logic a bit, removing toggleUsePriorityFees to make things simpler. If you want manual fees, just set Exchange.updatePriorityFee() to be nonzero. If you want automatic fees, just call Exchange.toggleAutoPriorityFee().

### DIFF
--- a/src/exchange.ts
+++ b/src/exchange.ts
@@ -30,7 +30,6 @@ import { Asset } from "./assets";
 import { SubExchange } from "./subexchange";
 import * as instructions from "./program-instructions";
 import { Orderbook } from "./serum/market";
-import fetch from "node-fetch";
 
 export class Exchange {
   /**
--- a/src/exchange.ts
+++ b/src/exchange.ts
@@ -30,6 +30,7 @@ import { Asset } from "./assets";
 import { SubExchange } from "./subexchange";
 import * as instructions from "./program-instructions";
 import { Orderbook } from "./serum/market";
+import fetch from "node-fetch";
 
 export class Exchange {
   /**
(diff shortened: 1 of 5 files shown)


src/exchange.ts
```diff
@@ -617,6 +612,41 @@ export class Exchange {
...
+        )}"]]}`,
+        headers: { "Content-Type": "application/json" },
+      });
+
+      let fees = (await data.json()).result
+        .sort((a, b) => b.slot - a.slot) // Sort descending
+        .slice(0, 20) // Grab the latest 20
+        .map((obj) => obj.prioritizationFee); // Take a list of prioritizationFee values only
+
+      let median = utils.median(fees);
```
filipzeta, line 640: Is median the best measure here? it's what Jupiter uses too. Here's some example medians from running it locally: 
image
The occassional 20k fees could skew the average quite a bit, but I'm open to any opinions here

xjcaa: seems ok to me :)

src/exchange.ts
```diff
@@ -672,6 +702,9 @@ export class Exchange {
                 await subExchange.handlePolling(callback);
               })
             );
+            if (this.useAutoPriorityFee == true) {
+              await this.updateAutoFee();
```
xjcaa, line 706: this is so frequent, could you change the cadence?

xjcaa: it's like 2-3x per second

src/exchange.ts
```diff
@@ -30,6 +30,7 @@ import { Asset } from "./assets";
 import { SubExchange } from "./subexchange";
 import * as instructions from "./program-instructions";
 import { Orderbook } from "./serum/market";
+import fetch from "node-fetch";
```
MeanMangosteen, line 33: Since you're designing this library for both node and browser environments, try to use 'universal' javascript/ts libraries. 

`node-fetch` won't exist in the browser environment. You could use sth like `axios` or similar, which run in both envs by design.

src/exchange.ts
filipzeta: wait shouldnt it be 30s? because of the check on line 695-697 with `this._pollInterval`, which defaults to 30s

filipzeta: yea its 30s, checked it locally
image

src/exchange.ts
filipzeta: removed import as per our chat - seems like just using fetch() is the right thing

MeanMangosteen (approved):

PR #244, opened 2023-06-28
Post-CMA improvements

filipzeta:
A couple of extra bits:
- improve MA -> CMA migration
- extra backend bits in IDL

### DIFF
--- a/src/cross-client.ts
+++ b/src/cross-client.ts
@@ -616,11 +616,38 @@ export class CrossClient {
     }
   }
 
-  public async migrateToCrossMarginAccount(
-    marginAccounts: PublicKey[]
-  ): Promise<TransactionSignature[]> {
+  public async findUserMarginAccounts(): Promise<PublicKey[]> {
+    let marginAccounts = [];
+
+    await Promise.all(
+      Exchange.assets.map(async (asset) => {
+        // Address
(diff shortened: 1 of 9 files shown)


src/cross-client.ts
```diff
@@ -616,12 +616,39 @@ export class CrossClient {
...
+          Exchange.programId,
+          Exchange.getZetaGroupAddress(asset),
+          this.publicKey
+        );
+
+        // Check if the address is valid
+        let account =
+          Exchange.program.account.marginAccount.fetchNullable(address);
+
+        if (account != null) {
```
kelvinlau20100, line 635: nit; can't this just be `if (account)`

src/risk-utils.ts
```diff
@@ -1,5 +1,5 @@
 import { BN } from "@zetamarkets/anchor";
-import { types, Exchange, constants, assets, instructions } from ".";
+import { types, Exchange, constants, assets, instructions, utils } from ".";
```
kelvinlau20100, line 2: is this import needed no other code was added?

filipzeta: removed

src/cross-client.ts
filipzeta: fixed

kelvinlau20100 (approved):

PR #245, opened 2023-06-29
Force cancel improvements for CrossMarginAccount

filipzeta:
Support both `marginAccount` and `crossMarginAccount` in `forceCancelOrderByOrderId()`  and `forceCancelOrders()`

### DIFF
--- a/src/cross-client.ts
+++ b/src/cross-client.ts
@@ -616,8 +616,12 @@ export class CrossClient {
     }
   }
 
-  public async findUserMarginAccounts(): Promise<PublicKey[]> {
-    let marginAccounts = [];
+  public async findUserMarginAccounts(): Promise<{
+    addresses: PublicKey[];
+    accounts: programTypes.MarginAccount[];
+  }> {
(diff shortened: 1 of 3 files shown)


src/cross-client.ts
```diff
@@ -1609,28 +1609,45 @@ export class CrossClient {
...
     orderId: anchor.BN,
     side: types.Side
   ): Promise<TransactionSignature> {
     this.delegatedCheck();
 
-    // TODO this should support margin + cross margin automatically
-    let account = (await Exchange.program.account.crossMarginAccount.fetch(
-      crossMarginAccountToCancel
-    )) as unknown as CrossMarginAccount;
+    let account =
```
kelvinlau20100, line 1618: i don't like this, i would prefer if we only did one fetch

filipzeta: cleaned up to only do one fetch. ran liquidation_perps.ts integration test locally and it's all fine too.

kelvinlau20100 (approved):

PR #246, opened 2023-06-30
Example cleanup

filipzeta:
Clean up examples to make sure they run with latest SDK changes.


examples/cranking/cranking-example.ts
```diff
@@ -32,7 +32,7 @@ const NETWORK_URL = process.env["network_url"]!;
 const MAX_ACCOUNTS_TO_FETCH = 99;
 let crankingMarkets = new Array(constants.ACTIVE_MARKETS - 1).fill(false);
 console.log(NETWORK_URL);
-const assetList = [assets.Asset.BTC];
+const assetList = [constants.Asset.BTC];
```
kelvinlau20100, line 35: is this asset list still relevant?

kelvinlau20100 (approved):

examples/cranking/cranking-example.ts
filipzeta: eh kinda because you can still update pricing and crank for a single market

PR #247, opened 2023-07-05
Add cross margin account state for per asset distinction

kelvinlau20100:


src/types.ts
```diff
@@ -209,6 +209,24 @@ export interface MarginAccountState {
...
 
+export interface CrossMarginAccountState {
+  balance: number;
+  availableBalanceInitial: number;
+  availableBalanceMaintenance: number;
+  availableBalanceWithdrawable: number;
+  state: Map<
+    Asset,
+    {
+      balance: number;
```
xjcaa, line 220: what dis

src/risk-utils.ts
```diff
@@ -1,17 +1,42 @@
...
   ProductLedger,
 } from "./program-types";
-import {
-  convertNativeBNToDecimal,
-  convertDecimalToNativeInteger,
-  getProductLedger,
-} from "./utils";
+import { getProductLedger } from "./utils";
+
+export function collectIteratorSum(x: IterableIterator<number>) {
```
xjcaa, line 12: Feel like this function doesn't need to exist? you're only using it on a map value right

```
Array.from(mapVariable.values()).reduce((a, b) => a + b, 0)
```

xjcaa (approved):
Add changelog

PR #250, opened 2023-07-18
add header to event queue fetch

kelvinlau20100:


filipzeta (approved):
LGTM, i reckon we can just release this as 1.4.5 when its deployed

PR #251, opened 2023-07-21
Trigger orders (StopLoss/TakeProfit)

filipzeta:
This PR implements trigger orders, which will be utilised for stop-loss/take-profit functionality. 

Functionality defaults to `FillOrKill` with trigger direction acting as a stop-loss/take-profit order (`INCREASING` if Ask, `DECREASING` if Bid), but you can use any order type and direction you wish to create more complicated things such as stop-limit orders.

Trigger price and order price are separate - as we don't have market order functionality in the program it's probably best to do `orderPrice = triggerPrice + slippage`, unless manually specified otherwise.

Trigger orders expire after `State.triggerOrderTimeoutSeconds`, and they can be deleted by an admin or the owner. They can also be edited by the owner (just the asset cannot be changed).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ Version changes are pinned to SDK releases.
 
 ## Unreleased
 
+## [1.8.0] 2023-08-29
+
+- New feature: Trigger Orders. ([#251](https://github.com/zetamarkets/sdk/pull/251))
+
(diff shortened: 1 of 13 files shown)


filipzeta:
- [ ] Changelog and SDK version before merging

kelvinlau20100 (approved):

PR #252, opened 2023-07-24
getMaxTradeSize(), getLiquidationPrice() and uPnL overhaul

filipzeta:
- getMaxTradeSize() - figures out the maximum size you can trade based on if you're closing or opening a position.
- getLiquidationPrice() - figures out at what price your position will get liquidated.
- Split uPnL calcs into estimateRealizedPnl() and calculateUnrealizedPnl()
- checkLiquidity() - finds the best price to execute an order at

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -1714,7 +1714,7 @@ export function median(arr: number[]): number | undefined {
 export const checkLiquidity = (
   size: number,
   asset: Asset,
-  orderbook?: types.DepthOrderbook
+  orderbook: types.DepthOrderbook
 ): types.LiquidityCheckInfo => {
   const side = size > 0 ? types.Side.BID : types.Side.ASK;
   // We default to min lot size to still show a price
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -1710,3 +1710,81 @@ export function median(arr: number[]): number | undefined {
   const mid = Math.floor(s.length / 2);
   return s.length % 2 === 0 ? (s[mid - 1] + s[mid]) / 2 : s[mid];
 }
+
+export const checkLiquidity = (
+  size: number,
+  asset: Asset,
+  orderbook?: types.DepthOrderbook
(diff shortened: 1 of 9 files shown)


MeanMangosteen (approved):

filipzeta:
- [ ] Split uPnL calc with executionPrices into a separate function for estimated realized PnL

src/utils.ts
```diff
@@ -1710,3 +1710,81 @@ export function median(arr: number[]): number | undefined {
...
+  // Invalid order
+  if (!orderbookSide || !orderbookSide.length) {
+    return { validLiquidity: false, avgPrice: 0, worstPrice: 0 };
+  }
+
+  let validLiquidity = false;
+  // Size seen on the orderbook that satisfies the specified price
+  let seenSize = 0;
+  // The cumulative trade value for the seen size
+  let cumAmount = 0;
```
MeanMangosteen, line 1734: I find this highly inappropriate.

src/utils.ts
```diff
@@ -1710,3 +1710,81 @@ export function median(arr: number[]): number | undefined {
   const mid = Math.floor(s.length / 2);
   return s.length % 2 === 0 ? (s[mid - 1] + s[mid]) / 2 : s[mid];
 }
+
+export const checkLiquidity = (
+  size: number,
+  asset: Asset,
+  orderbook?: types.DepthOrderbook
```
MeanMangosteen, line 1717: Given the logic, it makes more sense to make this required now. It's needed to get the liquidity whereas in the FE there was a fallback value for orderbook if it wasn't provided.

src/utils.ts
filipzeta: 🤓

src/utils.ts
filipzeta: yep good point, ive set it to just be `orderbook:` now

MeanMangosteen (approved):

PR #257, opened 2023-08-21
Post-trade calcs

filipzeta:
Useful for FE to show a before/after for key account statistics.

These functions work by cloning the provided account and then using existing risk functions. Cloning is quite slow because it requires a deep clone on a big object, so I've provided the ability to provide your own cloned account as I assume we're going to be calling multiple of these at the same time.

To call all 5 functions on my machine it takes ~0.6ms if no executionInfo is provided, ~1.5ms if executionInfo is provided and an account is cloned beforehand (included in time measurement) and ~5ms if an account has to be cloned in each function.

### DIFF
--- a/src/risk-utils.ts
+++ b/src/risk-utils.ts
@@ -324,11 +324,14 @@ export function addFakeTradeToAccount(
  * @param executionInfo A hypothetical trade. Object containing: asset (Asset), price (decimal USDC), size (signed decimal), isTaker (whether or not it trades for full size)
  * @returns
  */
-export function cloneAccountAndFakeTrade(
+export function fakeTrade(
   marginAccount: CrossMarginAccount,
+  clone: boolean,
   executionInfo?: types.ExecutionInfo
 ): CrossMarginAccount {
-  let account = cloneDeep(marginAccount) as CrossMarginAccount;
+  let account = clone
+    ? (cloneDeep(marginAccount) as CrossMarginAccount)
+    : marginAccount;
   addFakeTradeToAccount(
     account,
     executionInfo.isTaker,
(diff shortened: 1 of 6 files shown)


src/risk-utils.ts
```diff
@@ -135,3 +142,203 @@ export function checkMarginAccountMarginRequirement(
...
+ * Simulate adding an extra position/order into an existing CrossMarginAccount.
+ * This will change the account! Therefore do a deep clone first if you want a new account to simulate.
+ * @param marginAccount The CrossMarginAccount itself
+ * @param isTaker Whether or not the order crosses the orderbook in full and becomes a position
+ * @param asset The market on which we're trading
+ * @param side Bid or ask
+ * @param price The trade price, in decimal USDC
+ * @param size The trade size, in decimal USDC
+ */
+export function addFakeTradeToAccount(
```
filipzeta, line 156: This is the same code that was in `getMaxTradeSize()` in `risk.ts`, no changes were made

filipzeta:
If calling multiple of these functions you'll get a speedup by deep-cloning the account yourself beforehand. Example code:
```
let clonedAccount = utils.deepCloneCrossMarginAccount(client.account);
let equity = Exchange.riskCalculator.getEquity(
    clonedAccount,
    executionInfo,
    false
);
let buyingPower = Exchange.riskCalculator.getBuyingPower(
    clonedAccount,
    constants.Asset.SOL,
    executionInfo,
    false
);
let freeCollateral = Exchange.riskCalculator.getFreeCollateral(
    clonedAccount,
    executionInfo,
    false
);
let marginUsage = Exchange.riskCalculator.getMarginUsagePercent(
    clonedAccount,
    executionInfo,
    false
);
let leverage = Exchange.riskCalculator.getLeverage(
    clonedAccount,
    executionInfo,
    false
);
```

MeanMangosteen (approved):

PR #262, opened 2023-08-30
Subscribe to orderbooks and clean up some old code

filipzeta:
- Subscribe to orderbook changes instead of polling
- Delete a bunch of old options stuff from the market files

### DIFF
--- a/src/serum/scripts/market-gen.ts
+++ b/src/serum/scripts/market-gen.ts
@@ -63,9 +63,9 @@ function populateMarketStore(
   let subExchange = Exchange.getSubExchange(asset);
   let decodedSerumMarket: any;
   if (index == constants.PERP_INDEX) {
-    decodedSerumMarket = subExchange.markets.perpMarket.serumMarket.decoded;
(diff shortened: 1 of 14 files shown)


src/serum/scripts/market-gen.ts
```diff
@@ -63,9 +63,9 @@ function populateMarketStore(
   let subExchange = Exchange.getSubExchange(asset);
   let decodedSerumMarket: any;
   if (index == constants.PERP_INDEX) {
-    decodedSerumMarket = subExchange.markets.perpMarket.serumMarket.decoded;
+    decodedSerumMarket = subExchange.markets.market.serumMarket.decoded;
```
kelvinlau20100, line 66: small nit: we can remove nesting on this condition

kelvinlau20100 (approved):
one comment otherwise lgtm

src/serum/scripts/market-gen.ts
filipzeta: fixed 😊

PR #265, opened 2023-09-05
Cancel all trigger orders + place order

filipzeta:
Useful for FE when closing positions. Cancel all trigger orders for a given asset and then place one order on the same asset.

Tested locally in the following conditions:
- 55 trigger orders active -> completes in 6 txs
- 5 trigger orders active -> completes in 1 tx
- 0 trigger orders active -> completes in 1 tx

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -85,6 +85,7 @@ export const MAX_SETTLE_AND_CLOSE_PER_TX = 4;
 export const MAX_CANCELS_PER_TX = 3;
 export const MAX_CANCELS_PER_TX_LUT = 13;
 export const MAX_GREEK_UPDATES_PER_TX = 20;
+export const MAX_TRIGGER_CANCELS_PER_TX = 10;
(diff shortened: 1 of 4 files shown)


src/constants.ts
```diff
@@ -85,6 +85,7 @@ export const MAX_SETTLE_AND_CLOSE_PER_TX = 4;
 export const MAX_CANCELS_PER_TX = 3;
 export const MAX_CANCELS_PER_TX_LUT = 13;
 export const MAX_GREEK_UPDATES_PER_TX = 20;
+export const MAX_TRIGGER_CANCELS_PER_TX = 10;
```
filipzeta, line 88: Max 15 and fails at 16 locally. I set the limit lower because realistically 99.9% of people won't have >10 triggers on FE, and it gives a bit of buffer to prevent FE issues if accounts change or LUTs aren't updated or whatever else. 

If they have >10 triggers it'll just require 2 wallet clicks instead of 1 😊

griffinzeta (approved):

PR #270, opened 2023-09-12
Add slot to callback

filipzeta:


src/exchange.ts
```diff
@@ -803,21 +809,21 @@ export class Exchange {
   }
 
   private subscribePricing(
-    callback?: (asset: Asset, type: EventType, data: any) => void
+    callback?: (asset: Asset, type: EventType, slot: number, data: any) => void
   ) {
-    let eventEmitter = this._program.account.pricing.subscribe(
```
filipzeta, line 808: Had to change this one to use .onAccountChange() because the .subscribe() method didn't give back a slot

src/cross-client.ts
```diff
@@ -373,7 +384,7 @@ export class CrossClient {
         client._lastUpdateTimestamp = Exchange.clockTimestamp;
 
         if (callback !== undefined) {
-          callback(null, EventType.USER, {
+          callback(null, EventType.USER, context.slot, {
```
dm-sierra, line 387: Does somebody else using the SDK? Or we are OK to broke callback signature?

dm-sierra (approved):

kelvinlau20100 (approved):

src/cross-client.ts
filipzeta: yea market makers, but it'll only break when updating SDK versions so it's not an immediate issue

PR #275, opened 2023-09-19
Bump solana web3js to latest

0xtristan:
Chunked fetches in sdk seem to break on node18+ due to the issues outlined here:
https://github.com/solana-labs/solana-web3.js/issues/1418


0xtristan:
Also why do we pin the solana version? Seems reasonable to not expect breaking changes

filipzeta (approved):

kelvinlau20100 (approved):

0xtristan:
Bump to 1.87.6?
https://github.com/solana-labs/solana-web3.js/releases

PR #278, opened 2023-09-28
Reduce TIF log spam

filipzeta:


src/exchange.ts
```diff
@@ -641,6 +641,12 @@ export class Exchange {
       })
     );
 
+    await Promise.all(
```
filipzeta, line 644: Exchange.load() would always start up with epochStartTs=0 and then get an update later, I added this to the load process so it starts properly

src/subexchange.ts
```diff
@@ -195,9 +195,13 @@ export class SubExchange {
 
     await m.serumMarket.updateDecoded(Exchange.connection);
 
-    console.log(
-      `${assetToName(this.asset)} SubExchange perp Serum market refreshed`
-    );
+    // Can get spammy on non-mainnet if no one is placing orders
+    // because TIF epochs automatically roll over on new orders
```
filipzeta, line 199: devnet rn has really old epochs so its spammy

kelvinlau20100 (approved):

PR #290, opened 2023-11-01
ZSTD compression for RPC websocket connections 

filipzeta:
@solana/web3.js doesn't support different encodings so I made a fork: https://github.com/zetamarkets/zeta-solana-web3.js. It essentially just adds some code into connection.ts to seamlessly allow both base64 and base64+zstd encodings.

This PR allows for the use of a Connection object from that fork, and defaults everything to using `base64+zstd` for the encoding. CPU usage is minimally increased but network usage is 5-10x lower.

### DIFF
--- a/src/cross-client.ts
+++ b/src/cross-client.ts
@@ -22,15 +22,16 @@ import {
 } from "./program-types";
 import {
   PublicKey,
-  Connection,
   Transaction,
   TransactionSignature,
   AccountInfo,
   Context,
+  Connection,
   TransactionInstruction,
   ConfirmOptions,
   SYSVAR_CLOCK_PUBKEY,
 } from "@solana/web3.js";
+import { PublicKey as PublicKeyZstd } from "zeta-solana-web3";
 import * as types from "./types";
 import * as instructions from "./program-instructions";
 import { EventType } from "./events";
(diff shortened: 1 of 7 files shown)


src/exchange.ts
```diff
@@ -88,19 +92,19 @@ export class Exchange {
   public get provider(): anchor.AnchorProvider {
     return this._provider;
   }
-  public get connection(): Connection {
-    return this._provider.connection;
+  public get connection(): ConnectionZstd {
+    return this._provider.connection as unknown as ConnectionZstd;
```
filipzeta, line 96: this is a bit ugly but idk if there's a better way to handle it...

src/cross-client.ts
```diff
@@ -2090,7 +2091,7 @@ export class CrossClient {
     side: types.Side
   ): Promise<TransactionSignature> {
     let accountInfo = await Exchange.connection.getAccountInfo(
-      marginAccountToCancel
+      marginAccountToCancel as PublicKeyZstd
```
filipzeta, line 2094: same with this `as PublicKeyZstd`. i dont love it... 🤷

src/exchange.ts
filipzeta: also eventually web3.js will have this stuff (they're doing a rewrite currently) so it's not permanent :D

griffinzeta (approved):

PR #294, opened 2023-11-06
Round priority fees

filipzeta:
Saw some errors like `The number 262.5 cannot be converted to a BigInt because it is not an integer` because we had decimal prio fees. This PR adds rounding to both the auto fee we fetch and also the fee we actually send, in case someone puts a decimal fee manually.


kelvinlau20100 (approved):
looks good

PR #296, opened 2023-11-08
Add a new liquidation price calc

kelvinlau20100:


src/risk.ts
```diff
@@ -1054,6 +1055,77 @@ export class RiskCalculator {
...
+    const liqAssetPledger = cma.productLedgers[assets.assetToIndex(asset)];
+    const posSizeNative = liqAssetPledger.position.size.toNumber();
+    const nativeCot = liqAssetPledger.position.costOfTrades.toNumber();
+
+    if (posSizeNative == 0) {
+      return 0;
+    }
+
+    const posSize = convertNativeLotSizeToDecimal(posSizeNative);
+    const C = convertNativeBNToDecimal(
```
xjcaa, line 1108: Maintenance margin requirement or MMR?

src/risk.ts
```diff
@@ -1054,6 +1055,77 @@ export class RiskCalculator {
...
+      );
+      const denom = posSize * C - posSize;
+      return Math.max(0, num / denom);
+    } else {
+      const num = convertNativeIntegerToDecimal(
+        cma.balance.toNumber() + nativeCot + K
+      );
+      const denom = Math.abs(posSize) * C - posSize;
+      return Math.max(0, num / denom);
+    }
```
xjcaa, line 1126:
```
      let nativeCot = posSize > 0 ? nativeCot : -nativeCot;
      let num = balance + nativeCot + K;
      let denom = Math.abs(posSize) * C - posSize;
      return Math.max(0, num/denom);

xjcaa (approved):
rest lgtm if cleared up on readability

PR #307, opened 2023-11-22
Fix anchor.bn truncation, improve invalid asset behaviour and improve market refresh robustness

filipzeta:
A bundle of a few small fixes:
- Fix anchor.BN truncation in risk calcs: small accounts will truncate uPnL to 0, causing accuracy issues with trade size and liq price calcs
- Replace throw with an undefined return for invalid asset: throwing is a bit too aggressive, and caused some issues in FE
- Add check for Exchange.isInitialized when refreshing markets: if loading is super slow then the .isHalted() check can cause syntax errors because Exchange.state is undefined

### DIFF
--- a/src/assets.ts
+++ b/src/assets.ts
@@ -42,7 +42,7 @@ export function assetToName(asset: Asset): string | null {
   if (asset == Asset.PYTH) return "PYTH";
   if (asset == Asset.UNDEFINED) return "UNDEFINED";
   if (asset == null) return null; // Some things, like clock callbacks, are for all assets and return asset=null
-  throw Error("Invalid asset");
+  return undefined;
 }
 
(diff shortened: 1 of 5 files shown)


src/assets.ts
MeanMangosteen, line 1: Curious how we distinguish `undefined` from `Asset.UNDEFINED`. 

e.g. it would be good if for nameToAsset and indexToAsset, we return `Asset.UNDEFINED` instead of `undefined` if no match, so we're always dealing with a return type of type `Asset` instead of having to check for truthiness every time we call the asset.

MeanMangosteen (approved):

src/assets.ts
filipzeta: changed to have asset undefined wherever possible 😊

PR #317, opened 2023-12-09
Fix: Sentry errors

griffinzeta:
Fixes two errors with sentry

1) `subExchange` map missing value for a certain key sometimes
Screenshot 2023-12-09 at 1 33 52 AM

2) `MarginRequirement` can be `undefined` in addition to `null`
Screenshot 2023-12-09 at 1 35 55 AM

### DIFF
--- a/src/exchange.ts
+++ b/src/exchange.ts
@@ -691,13 +691,14 @@ export class Exchange {
   }
 
   public getSubExchange(asset: Asset): SubExchange {
-    try {
-      return this._subExchanges.get(asset);
-    } catch (_e) {
-      throw Error(
+    const subExchange = this._subExchanges.get(asset);
+    if (subExchange === undefined) {
+      throw new Error(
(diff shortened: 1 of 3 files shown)


src/exchange.ts
```diff
@@ -691,13 +691,14 @@ export class Exchange {
   }
 
   public getSubExchange(asset: Asset): SubExchange {
-    try {
```
griffinzeta, line 694: Shouldn't return an error unless the actual map is undefined (which I don't think is possible) so I removed the try/catch.

Need to test though

src/risk.ts
```diff
@@ -6,7 +6,6 @@ import {
   convertNativeBNToDecimal,
   convertNativeLotSizeToDecimal,
   convertDecimalToNativeInteger,
-  convertDecimalToNativeLotSize,
```
griffinzeta, line 9: removed unused imports

filipzeta (approved):

PR #318, opened 2023-12-11
Add configurable min-lot and tick size to State

filipzeta:


src/utils.ts
```diff
@@ -51,6 +51,41 @@ import cloneDeep from "lodash.clonedeep";
...
 
+export function getNativeTickSize(asset: Asset): number {
+  return 100;
+}
+
+export function getDecimalMinLotSize(asset: Asset): number {
+  return getNativeMinLotSize(asset) / 10 ** constants.POSITION_PRECISION;
+}
+
+export function getNativeMinLotSize(asset: Asset): number {
```
MeanMangosteen, line 62: Curious why this isn't an a record in constants vs a getter acting like a record.

kelvinlau20100: because it will require accessing the Exchange object in the future so it can't be static if that makes sense

xjcaa (approved):

src/utils.ts
MeanMangosteen: I see, makes sense.

src/utils.ts
```diff
@@ -51,6 +51,37 @@ import cloneDeep from "lodash.clonedeep";
 import * as os from "os";
 import { OpenOrders, _OPEN_ORDERS_LAYOUT_V2 } from "./serum/market";
 
+export function getNativeTickSize(asset: Asset): number {
+  return Exchange.state.tickSizes[assets.assetToIndex(asset)];
+}
+
+export function getDecimalMinLotSize(asset: Asset): number {
+  return getNativeMinLotSize(asset) / 10 ** constants.POSITION_PRECISION;
```
MeanMangosteen, line 59: could use `convertNativeLotSizeToDecimal`.

PR #326, opened 2023-12-18
Better prio fees

filipzeta:
Improve auto prio fees:
- Change calc to use max instead of median (toggleable)
- Significantly increase default upper limit

Example fees:
image


filipzeta:
I think FE would probably set low fees to be median, and then high fees to be max with _autoPriorityFeeOffset = 10k? probs no need for multiplier

kelvinlau20100 (approved):
update from main and add package.json and changelog

PR #330, opened 2023-12-21
Robustness improvements

filipzeta:
- Add maxRetries option to transactions, 0 is preferred for cranking to not spam on errors
- Split burn util up further to prevent it from eating RAM

### DIFF
--- a/src/risk.ts
+++ b/src/risk.ts
@@ -837,7 +837,7 @@ export class RiskCalculator {
     thresholdPercent: number = 1,
     bufferPercent: number = 5,
     maxIterations: number = 100
-  ): number {
+  ): number | undefined {
     if (tradePrice <= 0) {
       return undefined;
     }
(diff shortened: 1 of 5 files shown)


src/utils.ts
```diff
@@ -848,11 +848,17 @@ export function defaultCommitment(): ConfirmOptions {
...
 
 export function commitmentConfig(commitment: Commitment): ConfirmOptions {
-  return {
+  let opts = {
     skipPreflight: false,
     preflightCommitment: commitment,
     commitment,
   };
+
+  if (Exchange.maxRpcRetries != undefined) {
```
filipzeta, line 857: I've done it this way so that maxRetries in ConfirmOptions doesn't get set if you don't define Exchange.maxRpcRetries. Therefore it'll just do whatever the default is.

src/risk.ts
```diff
@@ -837,7 +837,7 @@ export class RiskCalculator {
     thresholdPercent: number = 1,
     bufferPercent: number = 5,
     maxIterations: number = 100
-  ): number {
+  ): number | undefined {
```
filipzeta, line 840: For FE syntax

xjcaa (approved):

PR #331, opened 2023-12-21
Fetch min lot size from state

xjcaa:


kelvinlau20100 (approved):
do we still need to set default value, aash didn't respond yet though, otherwise ookks fine

PR #335, opened 2023-12-27
Retry blockhash

filipzeta:
Retry the blockhash fetch 3 times

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -949,11 +949,30 @@ export async function processTransaction(
     );
   }
 
-  let recentBlockhash =
-    blockhash ??
-    (await provider.connection.getLatestBlockhash(
-      Exchange.blockhashCommitment
-    ));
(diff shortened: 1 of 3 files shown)


src/utils.ts
```diff
@@ -949,11 +949,30 @@ export async function processTransaction(
...
+  let recentBlockhash;
+  let blockhashRetryAmount = 0;
+  while (blockhashRetryAmount < 3) {
+    try {
+      recentBlockhash =
+        blockhash ??
+        (await provider.connection.getLatestBlockhash(
+          Exchange.blockhashCommitment
+        ));
+    } catch (e) {
```
filipzeta, line 961: Can't find a great way to test this locally - turning off my internet while its fetching blockhash (I added a `sleep(2000)`) only gets me 1 error when i turn my internet back on 😞 

Any ideas?

PR #338, opened 2024-01-09
Add fake cancel optionality to get max trade size

kelvinlau20100:


src/risk.ts
```diff
@@ -834,6 +835,7 @@ export class RiskCalculator {
     tradePrice: number,
     maxLeverage: number = -1,
     isTaker: boolean = true,
+    fakeCancel: types.Order = undefined,
```
dm-sierra, line 838: I think better to put the argument in the end - in case if old are in use somewhere, there will be an error

dm-sierra (approved):

PR #350, opened 2024-02-07
Refactor settleAndBurnVaultTokens

filipzeta:
`.all()` was failing a lot now that we have ~40k accounts, so instead just grab a random subset of accounts to burn tokens for


kelvinlau20100 (approved):
lgtm

PR #356, opened 2024-02-16
Improve cancelAllMarketOrders for big txs

filipzeta:
We have too many assets, can't just slap all of these into a single transaction.
- Don't send a tx if the user doesn't have an order on that asset
- Split up into N transactions instead of just sending one big one
- Little extra code change to return the tx sig from edit trigger order functions


src/constants.ts
```diff
@@ -123,6 +123,8 @@ export const CHAINLINK_PID: PublicKey = new PublicKey(
 export const MAX_SETTLE_AND_CLOSE_PER_TX = 4;
 export const MAX_CANCELS_PER_TX = 3;
 export const MAX_CANCELS_PER_TX_LUT = 13;
+export const MAX_PRUNE_CANCELS_PER_TX = 3;
+export const MAX_PRUNE_CANCELS_PER_TX_LUT = 5;
```
filipzeta, line 127: I tested these limits locally by just placing orders on every market and calling `cancelAllMarketOrders()`

xjcaa (approved):

PR #362, opened 2024-02-26
Add txRetryAmount and closeAllPositions

filipzeta:
- Add a retry option to utils.processTransaction() and use it in cross-client.ts
- Add new CrossClient function closeAllPositions() that force closes all positions using market orders


src/cross-client.ts
```diff
@@ -1283,6 +1319,64 @@ export class CrossClient {
...
+   * Close all positions using market orders
+   * @param orderPrices Manual override for what price to send, decimal format. If not provided, the price will be markPrice + 5% slippage
+   */
+  public async closeAllPositions(orderPrices?: Map<Asset, number>) {
+    let ixs = [];
+
+    for (var position of [...this._positions.values()].flat()) {
+      let side = position.size < 0 ? types.Side.BID : types.Side.ASK;
+
+      // 5% slippage unless a specific price is given
```
kelvinlau20100, line 1332: i think we shouldn't default to this, lets force the arg to be defined also 5% is really huge slippage

filipzeta: yea fair, ive made it force orderPrices

kelvinlau20100 (approved):

PR #375, opened 2024-03-27
Better rounding for conversion utils

filipzeta:


src/utils.ts
```diff
@@ -772,11 +769,8 @@ export function convertDecimalToNativeLotSize(
   roundingFactor: number = constants.MIN_NATIVE_MIN_LOT_SIZE
 ): number {
   return (
-    parseInt(
-      (
-        (amount * Math.pow(10, constants.POSITION_PRECISION)) /
-        roundingFactor
-      ).toFixed(0)
+    Math.trunc(
+      (amount * Math.pow(10, constants.POSITION_PRECISION)) / roundingFactor
```
MeanMangosteen, line 773: Just to be super safe I would suggest using your bn library here to do all the arithmetic. Only because we are doing a trunc() and that be be dangerous with the whole floating point imprecision thing.

this is a contrived example but:
image
Trunc here would give 2*10^23 instead of 3*10^23.

With real world crazy decimals this can also happen with numbers in the order of magnitude relevant to our platform.

MeanMangosteen: > Trunc here would give 2 * 10^23 instead of 3 * 10^23.

This may not be truly actually, but the point being if we get 1.999... instead of 2, then the trunc() would give a v diff value.

MeanMangosteen (approved):

PR #381, opened 2024-04-07
Improve tx confirmation rate by spamming sendRawTransaction()

filipzeta:


src/utils.ts
```diff
@@ -1036,13 +1036,20 @@ export async function processTransaction(
 
       txSig = await provider.connection.sendRawTransaction(rawTx, {
         skipPreflight: true,
+        preflightCommitment: provider.connection.commitment,
```
filipzeta, line 1039: saw this one in a tweet, not sure if it helps but it cant hurt because we skip preflight

src/utils.ts
```diff
@@ -1064,7 +1071,7 @@ export async function processTransaction(
               return txSig;
             }
           }
-          await sleep(1500); // Don't spam the RPC
+          await sleep(500); // Don't spam the RPC
```
filipzeta, line 1074: note: we still wait for sendRawTransaction+getSignatureStatus+getBlockHeight so this is more like 1s total loop time - just wanted to avoid someone spamming too rapidly if their latency to the RPC is low double digits

MeanMangosteen (approved):

PR #385, opened 2024-04-15
jto

kelvinlau20100:


filipzeta (approved):
Looks good - give it a non-beta version and add changelog please 😊 

Two small things:
1. Should we have an automated tip for FE? Maybe something based on the current prio fee? Could also grab the median tips from some API
2. Should we send bundles instead of just a single tx? To save costs, or allow for higher tips

PR #398, opened 2024-05-14
Send to Jito /transactions and double down RPC endpoints

filipzeta:


src/utils.ts
```diff
@@ -1281,21 +1310,37 @@ export async function processTransaction(
...
+          jsonrpc: "2.0",
+          id: 1,
+          method: "sendTransaction",
+          params: [encodedTx],
+        };
+        promises.push(sendJitoTxCaught(payload));
+      }
+
+      // First promise is the main RPC
+      let txSig = (await Promise.all(promises))[0];
```
griffinzeta, line 1332: I think you might want to `Promise.race` here

filipzeta: hmmmm i think we do promise.all() for this first one because we need the tx sig, but then can do promise.race() for the rest probably?

kelvinlau20100 (approved):

PR #399, opened 2024-05-16
Update subscribePricing

griffinzeta:
* Set the return value of the callback to an object mapping assets to their mark prices


filipzeta:
ok full stack king slay

filipzeta (approved):
looks good, ran it locally and it worked fine

PR #401, opened 2024-05-27
Optimize load

kelvinlau20100:


src/market.ts
```diff
@@ -73,39 +73,26 @@ export class ZetaGroupMarkets {
...
-      );
-      serumMarket = SerumMarket.loadFromDecoded(
-        decoded,
-        {
-          commitment: opts.commitment,
-          skipPreflight: opts.skipPreflight,
-        },
-        constants.DEX_PID[Exchange.network] as PublicKeyZstd
-      );
-    } else {
```
filipzeta, line 98: does removing this wreck new asset launches? because we have to load without store for a bit i think

filipzeta (approved):

PR #414, opened 2024-08-26
fix 'Invalid account discriminator' error when close account

frieadgood:
add the try and catch to ignore `Error: Invalid account discriminator` when closing the cross margin account and account manager.


filipzeta:
Looks good ty 😊 can you please bump the version in package.json and also add a new line in changelog.md (you might need to merge master again)? i can release it, just good to have those two

filipzeta (approved):

